PR #3, opened 2022-02-12
Solution

Abnormaltype:


denyslipin (approved):

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
 ignore = E203, E266, W503
 max-line-length = 79
-max-complexity = 18
+max-complexity = 19
```
IvanRamyk, line 4: Why change this?

IvanRamyk (changes requested):

.flake8
Abnormaltype: I've fixed

IvanRamyk (approved):
Good job!

PR #5, opened 2022-06-07
Solution

rotsen18:


app/main.py
```diff
@@ -1,34 +1,108 @@
...
+                self.decks.append(Deck(start[0], column))
+        else:
+            for row in (range(start[0], end[0] + 1)):
+                self.decks.append(Deck(start[1], row))
+
+    def _create_all_decks(self, coordinates):
+        self.decks = [
+            Deck(row_index, col_index)
+            for row_index, col_index in coordinates
+        ]
```
Magus6784, line 26: You never used it

app/main.py
```diff
@@ -1,34 +1,108 @@
...
+                for col_delta in range(-1, 2):
+                    check_coords = (row + row_delta, col + col_delta)
+                    ship_coords = [
+                        (deck.row, deck.col)
+                        for deck in ship.decks
+                    ]
+                    if (check_coords in self.field
+                            and check_coords not in ship_coords):
+                        raise Exception("Ships shouldn't be located"
+                                        "in the neighboring cells")
```
Magus6784, line 93: Better create specific error and don't use Exception

Magus6784 (changes requested):

app/main.py
rotsen18: Missed it after refactoring

app/main.py
rotsen18: If I were writing a full implementation of the game, I would create a new BattleError parent class with all kinds of exceptions. But here I decided enough basic exception and description to it.

Magus6784 (approved):

app/main.py
```diff
@@ -19,12 +19,6 @@ def _create_ship(self, start, end):
             for row in (range(start[0], end[0] + 1)):
                 self.decks.append(Deck(start[1], row))
```
DmytroSvirsa, line 21: maybe Deck(start[1], row) should be Deck(row, start[1])? it passes tests, but if you print the field, you will see, it's wrong

DmytroSvirsa (changes requested):

TPolina (approved):

PR #9, opened 2022-06-10
Solution

Serhii-Leonenko:


app/main.py
```diff
@@ -1,34 +1,106 @@
+class Field:
+    def __init__(self, added_ships):
+        self.field = []
+        for _ in range(10):
+            self.field.append(
```
KatiukhaO, line 5: i think, good idea, use class Field, it nice way

KatiukhaO (approved):

TPolina (approved):

PR #13, opened 2022-06-10
Solution first part main part of task

KatiukhaO:


app/main.py
```diff
@@ -1,34 +1,115 @@
...
 
 
 class Battleship:
-    def __init__(self, ships):
+
+    def __init__(self, ships: List[tuple]):
         # Create a dict `self.field`.
         # Its keys are tuples - the coordinates of the non-empty cells,
         # A value for each cell is a reference to the ship
         # which is located in it
```
rotsen18, line 13: If you want to describe the behavior of methods or classes, I would recommend using docstring. Then you need to change the description that the method should do something. He just does it.

rotsen18 (approved):

TPolina (approved):

PR #39, opened 2022-08-31
'Solution'

Diliaver:


app/main.py
```diff
@@ -1,34 +1,79 @@
...
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships, field={}):
+        self.ships = ships
+        self.field = field
+        self.field = Battleship.create(self)
+
+    def create(self):
```
IvanRamyk, line 57:
```suggestion
    def create_field(self):
```
Using the verb `create` isn't enough to describe what this method actually does

app/main.py
```diff
@@ -1,34 +1,79 @@
...
+        list_ships = []
+        for ships in self.ships:
+            battle_ships = list(ships)
```
IvanRamyk, line 60: Why convert a ship to the list?

app/main.py
```diff
@@ -1,34 +1,79 @@
...
+            list_ships.append(Ship(battle_ships[0], battle_ships[1]))
+        for create_decks in list_ships:
```
IvanRamyk, line 62:
```suggestion
        for ship in list_ships:
```
use nouns to name variables. it must be obvious what is stored inside

app/main.py
```diff
@@ -1,34 +1,79 @@
...
+            create_decks.decks_create()
+        for crate_field in list_ships:
```
IvanRamyk, line 64: same issue

app/main.py
```diff
@@ -1,34 +1,79 @@
...
+            for j in range(len(crate_field.decks) - 1):
+                self.field.update({(crate_field.decks[j].row,
```
IvanRamyk, line 66:
```suggestion
                self.field.update[(crate_field.decks[j].row, crate_field.decks[j].column] = crate_field
```
This way of creating a pair in dict is more compact

app/main.py
```diff
@@ -1,34 +1,79 @@
...
     def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+
+    def decks_create(self):
+        list_start = list(self.start)
```
IvanRamyk, line 16: It looks like useless converting

IvanRamyk (changes requested):

app/main.py
Diliaver: TypeError: 'builtin_function_or_method' object does not support item assignment

IvanRamyk: Sorry, forgot to delete `update` word:
```
 self.field[(crate_field.decks[j].row, crate_field.decks[j].column] = crate_field
```

IvanRamyk (approved):
Good job!

PR #56, opened 2022-09-27
Solution

KatyaVasylieva:


app/main.py
```diff
@@ -1,34 +1,139 @@
...
+
+        self.decks = [Deck(*start)]
+        self.is_drowned = is_drowned
+
+        # adding all ship decks to list:
+        # horizontal ship location
+        if start[0] == end[0]:
+
+            # few-decked ships
+            if start[1] != end[1]:
```
DmytroSvirsa, line 19: try to rewrite without this check (find better way to set range paramets)

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):
Well done!

PR #58, opened 2022-09-30
Solution

wQuelS:


app/main.py
```diff
@@ -1,34 +1,90 @@
...
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
+    def __init__(self, start: tuple, end: tuple, is_drowned=False) -> None:
         # Create decks and save them to a list `self.decks`
-        pass
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        self.create_fleet()
```
mltdll, line 15: Wow, to create an entire fleet out of decks of just one ship! Truly amazing.

tests/test_main.py
```diff
@@ -28,3 +28,4 @@ def test_battleship():
     assert battle_ship.fire((4, 6)) == "Sunk!"
     assert battle_ship.fire((9, 5)) == "Sunk!"
     assert battle_ship.fire((9, 6)) == "Miss!"
+    assert battle_ship.fire((9, 6)) == "Miss!"
```
mltdll, line 31: Are you sure It's necessary?

mltdll (changes requested):

tests/test_main.py
wQuelS: Oops, that was made accidentally

mltdll (approved):

app/main.py
```diff
@@ -1,34 +1,90 @@
...
-        pass
+        self.get_deck(row, column).is_alive = False
+        self.check_if_drowned()
+
+    def check_if_drowned(self) -> None:
+        for deck in self.decks:
+            if deck.is_alive:
+                self.is_drowned = False
+                return
+            self.is_drowned = True
```
Nattalli, line 43: can you try to do the same with `any()` and list comprehension? That is not a mistake, your solution is good. But I think that it will be more interesting

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+                    if self.field[(row, col)].is_drowned:
+                        field.append(" x ")
+                        continue
+                    for deck in self.field[(row, col)].decks:
+                        if deck.row == row and deck.column == col:
+                            if not deck.is_alive:
+                                field.append(" * ")
+                            else:
+                                field.append(" □ ")
+            print(*field)
```
Nattalli, line 90: did you make it for yourself?

app/main.py
Nattalli: WDYT?

Nattalli (changes requested):
Left some comments

app/main.py
wQuelS: nice idea. That would really improve code style, i guess.

app/main.py
wQuelS: yeap)

PR #61, opened 2022-10-03
Solution

StasOkhrym:


app/main.py
```diff
@@ -1,34 +1,63 @@
...
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self,
+             row: int,
+             column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        for deck in self.decks:
+            if deck.is_alive:
+                break
+        else:
```
mltdll, line 44: Using `else` with `for` and `while` loops is neither very intuitive* nor commonly used. I would advise against using it. However, I may be wrong.

\* For an average Python developer without any background in low-level languages

mltdll (changes requested):

mltdll (approved):

app/main.py
```diff
@@ -1,34 +1,63 @@
...
+                 start: tuple,
+                 end: tuple,
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        if self.start[0] == self.end[0] and self.start[1] == self.end[1]:
+            self.decks = [Deck(self.start[0], self.start[1])]
+            return
```
Nattalli, line 22: I don't like this empty return, but it is not a mistake

app/main.py
```diff
@@ -1,34 +1,63 @@
...
-        pass
+    def fire(self,
+             row: int,
+             column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        for deck in self.decks:
+            if deck.is_alive:
+                self.is_drowned = False
+                return
+            self.is_drowned = True
```
Nattalli, line 45: The such syntax seems easier. WDYT?
```python
if not any([deck.is_alive for deck in self.decks]):
            self.is_drowned = True
```

Nattalli (changes requested):

app/main.py
Nattalli: you can use `elif` in the next lines, so here you will not need this `return`. WDYT?

Nattalli (changes requested):

app/main.py
StasOkhrym: I like this. Thank you for hint 😃

Nattalli (approved):
Excellent! 🌞

PR #62, opened 2022-10-05
Solution

AVyha:


mltdll (changes requested):
Wow. You implemented everything in a completely different way than the instructions suggested. I'm not sure if this means the task has to be rejected.

app/main.py
```diff
@@ -1,34 +1,68 @@
...
+            self.ship_drowned(ship.ship)
+            return "Sunk!"
+        return "Hit!"
+
+    def ship_drowned(self, ship: list) -> None:
+        for deck in ship:
+            self.field[deck.row][deck.column] = "x"
+
+
+class Battleship(Ship):
```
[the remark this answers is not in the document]
AVyha: If i remove this inherit i should change 3 methods in Ship( Maybe later i try to change but now i dont have idea how to easy rebuild this block :)

mltdll (approved):
I guess it works...

AVyha:
> I guess it works...

i think you are right that i need remove inherit, im just say that im very lazy to do it now xD

app/main.py
```diff
@@ -1,34 +1,176 @@
...
+                     for i in range(self.len_ship)]
+        self.is_drowned = is_drowned
+
+    def get_deck(
+        self,
+        row: int,
+        column: int
+    ) -> Deck | Ship:
+
+        for ship in self.ships:
```
Nattalli, line 39: I see only `self.ship` in `__init__()` method. Where the `self.ships` is created?

app/main.py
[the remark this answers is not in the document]
Nattalli: that is the truth, you may to fix it

app/main.py
```diff
@@ -1,34 +1,176 @@
...
+            2: 0,
+            3: 0,
+            4: 0
+        }
+        for ship in self.ships:
+            we_have[ship.len_ship] += 1
+
+        assert must_have == we_have, \
+            "We need 1 four-deck ship, 2 three-deck ships, " \
+            "3 double-deck ships and 4 single-deck ships"
```
Nattalli, line 140: did you add your own test? Or why do you need it? That is a hard-code I think

app/main.py
```diff
@@ -1,34 +1,176 @@
...
+    # print(
+    #     battle_ship.fire((8, 3)),  #hit
+    #     battle_ship.fire((8, 4)),  #hit
+    #     battle_ship.fire((8, 5)),  #hit
+    #     battle_ship.fire((8, 6)),  #sunk
+    #     battle_ship.fire((0, 1)),  #sunk
+    #     battle_ship.fire((9, 9)),  #miss
+    #     battle_ship.fire((4, 1))  #hit
+    # )
+
```
Nattalli, line 175: also you can remove it from commit

Nattalli (changes requested):

Nattalli (approved):
Excellent! 🌞

PR #63, opened 2022-10-06
solution

KirillDrago:


app/main.py
```diff
@@ -1,34 +1,63 @@
...
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
```
mltdll, line 50: This attribute doesn't seem to be used anywhere but the initializer. Don't you think it should be removed?

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Good job!

PR #67, opened 2022-10-09
Solution

olia-trofymchuk:


app/main.py
```diff
@@ -1,34 +1,51 @@
...
+        for ship in ships:
+            battle_ship = Ship(ship[0], ship[1])
+            for deck in battle_ship.decks:
+                self.field[(deck.row, deck.column)] = battle_ship
+
+    def fire(self, location: tuple) -> str:
+        if location not in self.field:
+            return "Miss!"
+
+        self.field[location].fire(location[0], location[1])
```
wQuelS, line 46: U can use unpack here to make code more readable. E.g. self.field[location].fire(*location)
Everything else is really good!

app/main.py
```diff
@@ -1,34 +1,51 @@
...
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+
+        for deck in self.decks:
```
wQuelS, line 26: also, i guess, it would be better design style to put this into separete method.
One function - one purpose.

wQuelS (changes requested):

app/main.py
olia-trofymchuk: I agree w u. Thank u!

wQuelS (approved):
Looks nice to me:)

app/main.py
```diff
@@ -1,34 +1,53 @@
...
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        self.check_if_drowned()
+
+    def check_if_drowned(self) -> None:
+        for deck in self.decks:
+            self.is_drowned = True
+            if deck.is_alive:
+                self.is_drowned = False
+                return
```
Nattalli, line 32: does this `return` needed here?

Nattalli (changes requested):

app/main.py
olia-trofymchuk: no:)))))

Nattalli (approved):
Good! 👍🏻

PR #68, opened 2022-10-09
Solution

IvanStored:


app/main.py
```diff
@@ -1,34 +1,47 @@
...
+            if row == deck.row and column == deck.column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        for deck in self.decks:
```
mltdll, line 26: I think it would be a better design choice to extract this into a separate method.

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Well done!

PR #72, opened 2022-10-11
Solution

pazyn404:


app/main.py
```diff
@@ -1,34 +1,55 @@
...
-class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+        if cell in self._fired_decks:
+            return "Miss!"
+
+        self._fired_decks.add(cell)
+        self._count_alive_decks -= 1
+        return {0: "Sunk!"}.get(self._count_alive_decks, "Hit!")
```
mltdll, line 14: image

An interesting approach.
I can see no advantages over ternary operator in this case, though.

app/main.py
```diff
@@ -1,34 +1,55 @@
...
-        # Find the corresponding deck in the list
-        pass
+    def deck_alive(self, row: int, column: int) -> bool:
+        cell = (row, column)
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        if cell in self._fired_decks:
```
mltdll, line 19: You can use ternary operator here.
```python
return cell not in self._fired_decks
```

mltdll (changes requested):

mltdll (approved):

Nattalli (approved):
Well done!

LiudmylaKulinchenko (changes requested):
Test

LiudmylaKulinchenko (approved):
Test

PR #73, opened 2022-10-12
Solution!

ketstap162:


app/main.py
```diff
@@ -1,34 +1,83 @@
...
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self,
+                 start: tuple[int],
+                 end: tuple[int],
+                 is_drowned: bool = False) -> None:
+        if not (start[0] == end[0] or start[1] == end[1]):
+            raise ValueError("Ships shouldn't be located "
```
mltdll, line 17: This error message doesn't correspond with the check performed. It will only be raised when is not one-dimensional (e.g. its width is not 1).

app/main.py
```diff
@@ -1,34 +1,83 @@
...
+        self.ships = {(x, y): Ship(x, y) for x, y in ships}
+        self.field = {}
+        self.field.update({deck: ship
+                           for ship in self.ships.values()
+                           for deck in ship.decks.keys()})
+        self._validate_field()
+
+    def fire(self, location: tuple) -> str:
+
+        if not self.field.get(location):
```
mltdll, line 48: I think it's better to use `in` here instead of `get()` method.

app/main.py
```diff
@@ -1,34 +1,83 @@
...
+                    continue
+                if not self.field[point].decks[point].is_alive:
+                    print("*\t", end="")
+                else:
+                    print(u"\u25A1" + "\t", end="")
+            print("\n")
+
+    def _validate_field(self) -> None:
+        # validate data is dict{decks: count of ship}
+        validate = {4: 1, 3: 2, 2: 3, 1: 4}
```
mltdll, line 76: Variable names are usually nouns, not verbs. I'd suggest to rename this one into something in the spirit `correct_ship_count`. The name provided is only for you to get the idea.

mltdll (changes requested):

app/main.py
```diff
@@ -1,34 +1,104 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.ships = {(x, y): Ship(x, y) for x, y in ships}
+        self.field = {}
+        self.field.update({deck: ship
+                           for ship in self.ships.values()
+                           for deck in ship.decks.keys()})
```
mltdll, line 43: Iterating over `dict.keys()` is the same as iterating over the dict.

mltdll (approved):

app/main.py
ketstap162: It's more readable when there is an indication that `decks` is a dictionary

PR #77, opened 2022-10-16
added checklist.md

Nattalli:


checklist.md
```diff
@@ -0,0 +1,78 @@
...
+    self.length += 2
+```
+
+Bad example:
+```python
+def change_item(self):
+    self.length += 2
+    return
+```
+
```
Abnormaltype, line 24: add type annotation

checklist.md
```diff
@@ -0,0 +1,78 @@
...
+    # do something
+    return
+if a == d:
+    # do something
+    return
+```
+
+### 3. Don't use redundant `elif`.
+
+Don't use redundant `elif` if a non-empty `return` is used:
```
Abnormaltype, line 53: where is an example with `elif`?

Abnormaltype (approved):

PR #82, opened 2022-10-26
solution

dkyrpan:


app/main.py
```diff
@@ -1,34 +1,51 @@
...
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
+        self.field = {}
+
+        for ship in ships:
+            battleship = Ship(ship[0], ship[1])
+            for deck in battleship.decks:
+                self.field[(deck.row, deck.column)] = battleship
```
Abnormaltype, line 43: create a function for this action

Abnormaltype (changes requested):

app/main.py
dkyrpan: This action is included in def __init__
Where exactly should the function be and what then to do with it? 
Now everything seems to work correctly, I do not understand ...

Abnormaltype (approved):
Good job!

PR #85, opened 2022-11-04
Solution

TetyanaPavlyuk:


app/main.py
```diff
@@ -1,34 +1,134 @@
...
+            for col in range(10):
+                if (row, col) not in self.field.keys():
+                    print("~", end=end_value)
+                else:
+                    for coord, ship in self.field.items():
+                        if (row, col) == coord:
+                            if ship.is_drowned:
+                                print("x", end=end_value)
+                            else:
+                                for deck in ship.decks:
```
IvanRamyk, line 90: `Flat is better than nested.` due to the zen of python

this code seems too nested. try to fix it

app/main.py
```diff
@@ -1,34 +1,134 @@
...
+                is_alive = True
+                break
+        if not is_alive:
+            self.is_drowned = True
 
 
 class Battleship:
-    def __init__(self, ships):
+    def __init__(self, ships: list) -> None:
         # Create a dict `self.field`.
```
IvanRamyk, line 47: remove these comments after implementing the corresponding method

IvanRamyk (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #101, opened 2022-12-15
Solution

IvanMayor:


app/main.py
```diff
@@ -1,34 +1,61 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        self.change_is_drowned()
+
+    def change_is_drowned(self) -> None:
+        if any(dec.is_alive for dec in self.decks):
+            self.is_drowned = False
+        else:
+            self.is_drowned = True
```
IvanRamyk, line 44: could be simplified to one line 
```suggestion
        self.is_drowned = not any(dec.is_alive for dec in self.decks)
```

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,34 +1,58 @@
...
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self, start: tuple,
+                 end: tuple,
```
ArsenPidhoretskyi: tuple of what add inner type

app/main.py
```diff
@@ -1,34 +1,58 @@
...
+        if self.start[0] == self.end[0] and self.start[1] == self.end[1]:
+            self.decks.append(Deck(self.start[0], self.end[1]))
+
+        elif self.start[0] != self.end[0]:
+            for i in range(self.start[0], self.end[0] + 1):
+                self.decks.append(Deck(i, self.start[0]))
+
+        elif self.start[1] != self.end[1]:
+            for i in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(self.start[0], i))
```
ArsenPidhoretskyi: Simplify code by usage loop inside loop

app/main.py
```diff
@@ -1,34 +1,58 @@
...
+
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,58 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        self.change_is_drowned()
+
+    def change_is_drowned(self) -> None:
+        self.is_drowned = not any(dec.is_alive for dec in self.decks)
```
```suggestion
        self.is_drowned = not any(deck.is_alive for deck in self.decks)
```

app/main.py
```diff
@@ -1,34 +1,58 @@
...
-        # which is located in it
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
```

ArsenPidhoretskyi (changes requested):

app/main.py
YukihiroSM: Not yet fixed.

app/main.py
```diff
@@ -1,34 +1,61 @@
...
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.field = {}
+        for ship in ships:
+            boat = Ship(ship[0], ship[1])
+            for deck in boat.decks:
+                self.field[(deck.row, deck.column)] = boat
+
+    def fire(self, location: tuple) -> str:
+        if location not in self.field.keys():
```
YukihiroSM: It's not necessary to use keys() here, because `in self.field` is using keys only by default.

YukihiroSM (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #102, opened 2022-12-15
Solution

ghost:


app/main.py
```diff
@@ -1,34 +1,63 @@
...
+        for row in range(start[0], end[0] + 1):
+            for col in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row, col))
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        return list(
+            filter(
```
IvanRamyk, line 20: use list comprehension instead of `filter`. it's more pythonic way

IvanRamyk (changes requested):
everything else is fine

OleksandrPozden (approved):

PR #108, opened 2023-01-03
Solution

derkach55:


app/main.py
```diff
@@ -1,34 +1,52 @@
...
+        if self.start[0] == self.end[0]:
+            self.decks = [
+                Deck(self.start[0], i)
+                for i in range(self.start[1], self.end[1] + 1)
+            ]
+        elif self.start[1] == self.end[1]:
+            self.decks = [
+                Deck(i, self.start[0])
+                for i in range(self.start[0], self.end[0] + 1)
+            ]
```
katevasilenko, line 23: For me it’s looks too heavy for a constructor. I would prefer to separate this logic to a method or do some refactor to make it looks a bit compact and readable

katevasilenko (approved):
Looks good, but would be great if do some refactoring

PR #113, opened 2023-01-13
Solution

mish-sk:


app/main.py
```diff
@@ -1,34 +1,48 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        for deck in self.decks:
+            self.is_drowned = True
+            if deck.is_alive is True:
+                self.is_drowned = False
```
Serhii-Leonenko, line 31: You don't need to change self.is_drowned parameter on every iteration, use all() with comprehension to check if all decks are not alive, if all() is True -> self.is_drowned = True

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,34 +1,46 @@
...
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self,
+                 start: tuple,
+                 end: tuple,
+                 is_drowned: bool = False) -> None:
```
andrewbelo, line 12:
```suggestion
    def __init__(
        self,
                 start: tuple,
                 end: tuple,
                 is_drowned: bool = False
      ) -> None:
```

andrewbelo (changes requested):

Serhii-Leonenko (approved):
GJ!

PR #118, opened 2023-01-19
Solution

evgenijmartynuk07:


app/main.py
```diff
@@ -1,34 +1,151 @@
...
+        length_ships = [self.field[ship].length for ship in self.field]
+        if length_ships.count(4) != 1:
+            raise Exception("there should be 1 four-deck ship")
+        if length_ships.count(3) != 2:
+            raise Exception("there should be 2 three-deck ships")
+        if length_ships.count(2) != 3:
+            raise Exception("there should be 3 double-deck ships")
+        if length_ships.count(1) != 4:
+            raise Exception("there should be 4 single-deck ships")
+        neighboring_cells = [ship for ship in self.field]
```
ArsenPidhoretskyi, line 89: It coordinates not ship
also you can use list(self.field.keys())

app/main.py
```diff
@@ -1,34 +1,151 @@
...
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    field = {}
+
+    def __init__(self, ships: tuple) -> None:
```
ArsenPidhoretskyi, line 72: provide full type for parameter ships

app/main.py
```diff
@@ -1,34 +1,151 @@
...
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    decks = []
```
ArsenPidhoretskyi, line 16: in must be instance property, not class

ArsenPidhoretskyi (changes requested):
1. Provide typing for all variables(full typing)
2. Correct name variables, so hard to read code
3. All attributes must be instance not class 

Check that in all places, not only where I left comments. After I will review again)

app/main.py
```diff
@@ -1,34 +1,151 @@
...
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.length = None
+        self.length_of_ship()
```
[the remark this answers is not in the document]
evgenijmartynuk07: came up with to understand that the ship went to the bottom)))

Vasyl-Smutok (approved):
Good Job 👍

app/main.py
```diff
@@ -1,34 +1,147 @@
...
+    ) -> None:
+
+        self.decks = []
+        self.length = None
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+
+    @property
+    def get_length(self) -> int:
```
ArsenPidhoretskyi, line 29: Property is a noun. Correct naming just length

app/main.py
```diff
@@ -1,34 +1,147 @@
...
-        pass
+    def __init__(
+            self,
+            start: tuple[int, int],
+            end: tuple[int, int],
+            is_drowned: bool = False
+    ) -> None:
+
+        self.decks = []
+        self.length = None
```
ArsenPidhoretskyi, line 23: Remove this. Just use property

app/main.py
```diff
@@ -1,34 +1,147 @@
...
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+
+    @property
+    def get_length(self) -> int:
+        self.length = max(
+            self.end[1] - self.start[1], self.end[0] - self.start[0]
+        ) + 1
+        return self.length
```
ArsenPidhoretskyi, line 33:
```suggestion
        return max(
            self.end[1] - self.start[1], self.end[0] - self.start[0]
        ) + 1
```

app/main.py
ArsenPidhoretskyi: You do not need to update state in properties also

app/main.py
```diff
@@ -1,34 +1,147 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    @property
```
ArsenPidhoretskyi, line 35: It is method)

app/main.py
```diff
@@ -1,34 +1,147 @@
...
+
+    def get_deck(self, row: int, column: int) -> str:
+        for deck in self.get_decks:
+            if deck.row == row and deck.column == column:
+                return self.fire(deck)
+
+    def fire(self, deck: Deck) -> str:
+        if self.is_drowned is False:
+            deck.is_alive = False
+            self.length -= 1
```
ArsenPidhoretskyi, line 54: You can use as property and it will be autocalculate)

ArsenPidhoretskyi (changes requested):
Please check the difference between method and property
method name must be verb
property name must be noun

PR #122, opened 2023-01-26
solution

IhorPokr:


app/main.py
```diff
@@ -1,34 +1,139 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        dead_deck = self.get_deck(row, column)
+        dead_deck.is_alive = False
+        is_at_least_one_deck_alive: bool = False
+        for deck in self.decks:
+            if deck.is_alive:
+                is_at_least_one_deck_alive = True
+        if not is_at_least_one_deck_alive:
+            self.is_drowned = True
```
YuriiKindrat, line 44: no need to go through all the list of decks u can make loop and if some deck.is alive, is_at_least_one_deck_alive = True and break the loop

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #123, opened 2023-01-28
Solution

YuriiVataschuk:


app/main.py
```diff
@@ -1,34 +1,68 @@
...
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        self.decks = []
+        for row in range(start[0], end[0] + 1):
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row, column))
```
ArsenPidhoretskyi, line 24: maybe in this case use list comprehension?

app/main.py
```diff
@@ -1,34 +1,68 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(
+            self,
+            row: int,
+            column: int
+    ) -> Deck:
```
ArsenPidhoretskyi, line 30: Due to your code it returns Optional[Deck]
When change do not forgot to check for None in all places

ArsenPidhoretskyi: Or if deck was not found raise an Error in this method

app/main.py
```diff
@@ -1,34 +1,68 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(
+            self,
+            row: int,
+            column: int
+    ) -> None:
+        self.get_deck(row, column).is_alive = False
```
ArsenPidhoretskyi, line 40: It will fail if deck was not found

app/main.py
```diff
@@ -1,34 +1,68 @@
...
+        for deck in self.decks:
+            self.is_drowned = True
+            if deck.is_alive:
+                self.is_drowned = False
+                break
```
ArsenPidhoretskyi, line 45: Not so good to set always self.is_drowned = True
move this action from the cycle
BTW, use any() function instead of loop

app/main.py
```diff
@@ -1,34 +1,68 @@
...
 
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def fire(
+            self,
+            location: tuple
```
ArsenPidhoretskyi, line 61:
```suggestion
            location: tuple[int, int]
```

app/main.py
```diff
@@ -1,34 +1,68 @@
...
 class Battleship:
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(
+            self,
+            ships: list
```
ArsenPidhoretskyi, line 51: list of what?

app/main.py
```diff
@@ -1,34 +1,68 @@
...
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
```
ArsenPidhoretskyi, line 17:
```suggestion
            start: tuple[int, int],
            end: tuple[int, int],
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #126, opened 2023-02-01
Develop

rakamakaphone:
partial solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,8 @@
 class Deck:
     def __init__(self, row, column, is_alive=True):
-        pass
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
--- a/app/main.py
+++ b/app/main.py
@@ -1,36 +1,103 @@
+from __future__ import annotations
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
+    def __init__(self, row: int, column: int,
+                 is_alive: bool = True) -> None:
         self.row = row
         self.column = column
         self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self, start: tuple, end: tuple,
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def create_deck(self) -> None:
+        if self.start[0] == self.end[0]:
+            for i in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(self.start[0], i))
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,34 +1,54 @@
...
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self, start: tuple, end: tuple,
+                 is_drowned: bool = False) -> None:
```
Bohdan-Salamakha, line 14: how can u create ship that already drowned? don't allow suer to do it
just self.is_drowned = False

app/main.py
```diff
@@ -1,34 +1,54 @@
+from __future__ import annotations
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(self, row: int, column: int,
+                 is_alive: bool = True) -> None:
```
Bohdan-Salamakha, line 6: when created, deck is always alive, doesn't it?
remove is_alive from parameters and just declare bellow self.is_alive = True

app/main.py
```diff
@@ -1,34 +1,54 @@
...
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {ship: Ship(ship[0], ship[1]) for ship in ships}
+        for key, value in self.field.items():
```
Bohdan-Salamakha, line 43: I see u don't use key, just replace it with _

app/main.py
```diff
@@ -1,34 +1,54 @@
...
+            value.create_deck()
+
+    def fire(self, location: tuple) -> str:
+        for key, value in self.field.items():
+            if location[0] in range(key[0][0], key[1][0] + 1)\
```
Bohdan-Salamakha, line 48: remove \
just
            if (location[0] in range(key[0][0], key[1][0] + 1)
                    and location[1] in range(key[0][1], key[1][1] + 1)):

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,34 +1,52 @@
...
+                self.decks.append(Deck(self.start[0], i))
+        else:
+            for i in range(self.start[0], self.end[0] + 1):
+                self.decks.append(Deck(i, self.start[1]))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi, line 26: It returns Optional[Deck]

ArsenPidhoretskyi (approved):
Good! But also check type annotations comment

Bohdan-Salamakha (approved):

PR #132, opened 2023-02-07
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1,34 +1,119 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        is_drowned = True
+        for deck in self.decks:
+            if deck.is_alive is True:
+                is_drowned = False
```
Grayder0152, line 55: after that you can `break` loop

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+        if is_drowned:
```
Grayder0152, line 56: Instead of 56 and 57 lines use `self.is_drowned = is_drowned `

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
Grayder0152, line 51: What if `get_deck` return `None`?

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+            print(string)
+        print("")
+
+    def _validate_field(self, ships: list) -> None:
+        if len(ships) != 10:
+            raise IncorrectNumberOfShips
+        number_of_ships = {1: 0, 2: 0, 3: 0, 4: 0}
+        for ship in ships:
+            if ship[0][0] != ship[1][0] and ship[0][1] != ship[1][1]:
+                raise IncorrectShipCoordinates
```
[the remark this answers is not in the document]
timyr4iq: I don't quite understand what exactly I should do

app/main.py
timyr4iq: this is not possible due to the implementation of the algorithm. in the Battleship class, in the self.field field, I enter only non-empty cells that are parts of ships. Further, before each get_deck call, I check if the given coordinates are in self.field

app/main.py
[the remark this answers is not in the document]
Grayder0152: `raise IncorrectShipCoordinates()`

Grayder0152 (approved):

PR #142, opened 2023-02-21
Solution

6aPa6aH:


app/main.py
```diff
@@ -1,34 +1,52 @@
...
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> tuple:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        fired_deck = self.get_deck(row, column)
+        if fired_deck:
```
Grayder0152, line 32: More correct use `if fired_deck is not None`

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,34 +1,57 @@
...
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def create_deck(self) -> None:
+        if self.start[0] == self.end[0]:
+            for i in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(self.start[0], i))
+        else:
+            for i in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(i, self.start[1]))
```
ArsenPidhoretskyi, line 28: Can do without if, just you loop inside loop

app/main.py
```diff
@@ -1,34 +1,57 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Any:
```
ArsenPidhoretskyi, line 30: It returns not any! You can say concrete item

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #143, opened 2023-02-22
1 commit

AndereLion:


manjustice (approved):

app/main.py
```diff
@@ -1,34 +1,128 @@
...
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        alive = False
+        for deck in self.decks:
+            if deck.is_alive:
+                alive = True
+        if not alive:
+            self.is_drowned = True
```
YuriiKindrat: it is to comlicated u can use any/all funcs with list comp. to check all decks

app/main.py
```diff
@@ -1,34 +1,128 @@
...
+    def decks_maker(start: tuple, end: tuple) -> list[Deck]:
+        decks = []
+        if start[0] - end[0] != 0:
+            for row in range(start[0], end[0] + 1):
+                decks.append(Deck(row, end[1]))
+        elif start[1] - end[1] != 0:
+            for column in range(start[1], end[1] + 1):
+                decks.append(Deck(start[0], column))
+        elif start == end:
+            decks.append(Deck(start[0], start[1]))
```
YuriiKindrat: u can use here only 2 conditions, try to solve this

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
```diff
@@ -1,34 +1,169 @@
...
+    ]
+)
+
+print(
+    battle_ship.fire((0, 4)),  # Miss!
+    battle_ship.fire((0, 3)),  # Hit!
+    battle_ship.fire((0, 2)),  # Hit!
+    battle_ship.fire((0, 1)),  # Hit!
+    battle_ship.fire((0, 0)),  # Sunk!
+)
```
YuriiKindrat: gj, but remove unnecessary code here

YuriiKindrat (changes requested):

app/main.py
AndereLion: Removed =)

YuriiKindrat (approved):

PR #144, opened 2023-02-22
Battleship solution

bihunva:


SafonovVladimir (approved):
GJ

app/main.py
```diff
@@ -1,34 +1,46 @@
...
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self, start: list[int], end: list[int], is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        self.decks = [Deck(row, column)
+                      for row in range(start[0], end[0] + 1)
+                      for column in range(start[1], end[1] + 1)]
```
wQuelS: In my opinion it would be better to separate the responsibility of creating a ship and creating its decks into different methods.
There are several ways to implement it. For example, you can make `decks` as an empty list in `__init__` and also add something like `self.create_ship()` which will call method for ship creation. You can also add `start` and `end` points to `__init__` and pass to that method as well

app/main.py
```diff
@@ -1,34 +1,46 @@
...
+        self.get_deck(row, column).is_alive = False
+        alive_decks = [deck.is_alive for deck in self.decks]
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        if any(alive_decks):
+            return "Hit!"
+        else:
```
wQuelS: this `else` seems to be redundant here

wQuelS (changes requested):

app/main.py
bihunva: thanks, good point

YuriiKindrat (approved):

PR #146, opened 2023-02-23
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -20,8 +20,9 @@ class Ship:
         self.start = start
         self.end = end
         self.is_drowned = is_drowned
-        self.decks = []
-        self._add_to_decks()
+        self.decks = [Deck(row, column)
+                      for row in range(start[0], end[0] + 1)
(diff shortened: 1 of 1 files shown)


MaybeTI (approved):

app/main.py
```diff
@@ -1,34 +1,85 @@
...
+        for ship in self.ships.values():
+            for deck in ship.decks:
+                if (deck.row, deck.column) == location:
+                    deck.is_alive = False
+                    ship.live_decks -= 1
+                    if ship.live_decks == 0:
+                        ship.is_drowned = True
+                        return "Sunk!"
+                    return "Hit!"
+        else:
```
YuriiKindrat: no need to use else here

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: Fixed

YuriiKindrat (approved):

PR #147, opened 2023-02-23
Solution

MaybeTI:


app/main.py
```diff
@@ -1,34 +1,103 @@
...
+    pass
+
+
+class MissingSomeShips(Exception):
+    pass
 
 
 class Battleship:
-    def __init__(self, ships):
+    battle_field = [["~", "~", "~", "~", "~", "~", "~", "~", "~", "~"],
```
NEkropo1: I know some people would say you can do it with [~] * 10, or by list comprehension or anything, but I like how it's visualized )
But I have question on why is it not an instance variable? They should be used only inside of instance, not in class, right?

app/main.py
```diff
@@ -1,34 +1,103 @@
...
-    def __init__(self, row, column, is_alive=True):
-        pass
+# class Deck:
+#     def __init__(self, row, column, is_alive=True):
+#         pass
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
+    decks = []
```
NEkropo1, line 2: Same which I said about field in class Battleship, didn't mention before) I understand overall idea, that every ship should have different Decks, but that sounds not any flexible, as we gonna refactor a lot if we'll want to add any new logic inside Ship instances.

app/main.py
```diff
@@ -1,34 +1,103 @@
...
         # Create decks and save them to a list `self.decks`
-        pass
+        self.deck = self.create_deck(start, end)
+        Ship.decks.append(self)
+
+    @staticmethod
+    def create_deck(start: tuple, end: tuple) -> list:
+        if start[0] == end[0]:
+            return [(start[0], point) for point in range(start[1], end[1] + 1)]
+
```
NEkropo1, line 16:
```suggestion
```
Sry for this, but it seems pretty simple, so why should you add this space, can't get it.

app/main.py
```diff
@@ -1,34 +1,103 @@
...
+            2: 0,
+            1: 0
+        }
+        if len(self.field) != 10:
+            raise NotEnoughtShips("Pleas enter correct number of ships")
+        for ship in Ship.decks:
+            dic_ships[len(ship.deck)] += 1
+        if dic_ships != check_dict:
+            raise MissingSomeShips("Please check the correctness"
+                                   " of the entered data")
```
NEkropo1: Like separating of checks.

NEkropo1 (changes requested):
Clean

app/main.py
```diff
@@ -1,34 +1,103 @@
...
+                ship.deck.remove(location)
+                return "Hit!"
+        return "Miss!"
+
+    @staticmethod
+    def create_field(ships: list) -> dict:
+        res = {}
+        for ship in ships:
+            res[ship] = Ship(start=ship[0], end=ship[1])
+        return res
```
NEkropo1: Pretty sure the task was to make dict, where key was tuple of coordinates, which makes whole code a lot more efficient, since you won't search through ship in finding coords, but hit coords directly... Try to refactor that, 'cause that's not according to readme, even though I understand overall logic

app/main.py
YuriiKindrat: @NEkropo1 is right, no need to hardcode a battlefield like this, use list comp. etc.

YuriiKindrat (changes requested):
also remove unnecessary comments in ur code

YuriiKindrat (changes requested):
comments is still in ur code

YuriiKindrat (approved):

PR #148, opened 2023-02-24
Solution

NEkropo1:
i h8 m4tr1x ];(

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -21,23 +21,26 @@ class Deck:
 
 class Ship:
     def __init__(
-        self, start: tuple[int, int], end: tuple[int, int], is_drowned=False
+        self,
+        start: tuple[int, int],
+        end: tuple[int, int],
+        is_drowned: bool = False
     ) -> None:
         self.length = 0
         self.decks = []
         self.create_from_tuples_list(start, end)
         self.is_drowned = is_drowned
 
-    def get_deck(self, coords: tuple[int, int]):
+    def get_deck(self, coords: tuple[int, int]) -> Deck:
         for deck in self.decks:
             if tuple([deck.row, deck.column]) == coords:
                 return deck
 
-    def sunk(self):
+    def sunk(self) -> None:
         for deck in self.decks:
             deck.appearance = "x"
 
-    def fire(self, coords: tuple[int, int]):
+    def fire(self, coords: tuple[int, int]) -> str:
         deck = self.get_deck(coords)
         deck.is_alive = False
         deck.appearance = "*"
--- a/app/main.py
+++ b/app/main.py
@@ -81,7 +81,9 @@ class Battleship:
     def _validate_field(self) -> None:
         ships = list(self.field.values())
         expected_counts = {0: 10, 1: 4, 2: 3, 3: 2, 4: 1}
-        counts = {i: sum(len(ship.decks) == i for ship in ships) for i in range(5)}
+        counts = {
+            i: sum(len(ship.decks) == i for ship in ships) for i in range(5)
+        }
         for i in range(5):
             if counts[i] == expected_counts[i]:
                 raise ShipCountError(
@@ -98,10 +100,11 @@ class Battleship:
                     and (0 <= col + j < 10)
                     and (i != 0 or j != 0)
                 ]
-                if any(neighbor in self.field for neighbor in neighbors):
-                    raise ValueError(
-                        "Ships should not be located in neighboring cells"
-                    )
+                for neighbor in neighbors:
+                    if neighbor in self.field:
+                        raise ValueError(
+                            "Ships should not be located in neighboring cells"
+                        )
 
     def fire(self, location: tuple) -> str | None:
         if location not in self.field:
--- a/app/main.py
+++ b/app/main.py
@@ -1,34 +1,134 @@
+from typing import Any
+
+
+class ShipCountError(Exception):
+    pass
+
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,34 +1,137 @@
...
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(
+        self,
+        row: int,
+        column: int,
+        is_alive: bool = False,
```
WDemidenko, line 13: Not sure about this, in the assignment was given initially is_alive: bool = True, which makes sense since when we have just created a Deck it is alive

WDemidenko (changes requested):

app/main.py
WDemidenko, line 16: Not sure about this, in the assignment was given initially is_alive: bool = True, which makes sense since when we have just created a Deck it is alive

app/main.py
```diff
@@ -1,34 +1,137 @@
...
+        self._map = [[Deck(i, j) for j in range(10)] for i in range(10)]
+        self.field = self._create_ships(ships)
+
+    def _validate_field(self) -> None:
+        ships = list(self.field.values())
+        expected_counts = {0: 10, 1: 4, 2: 3, 3: 2, 4: 1}
+        counts = {
+            i: len([s for s in ships if len(s.decks) == i]) for i in range(5)
+        }
+        for i in range(5):
```
WDemidenko, line 87: You are repeating the same loop several times, refactor the code to reduce the complexity of the algorithm and speed up the program. I am talking about loops "for i in range 5" and "for * in ships"

app/main.py
```diff
@@ -1,34 +1,137 @@
...
+        for ship in ships:
+            for deck in ship.decks:
+                row, col = deck.row, deck.column
+                neighbors = [
+                    (row + i, col + j)
+                    for i in range(-1, 2)
+                    for j in range(-1, 2)
+                    if (0 <= row + i < 10)
+                    and (0 <= col + j < 10)
+                    and (i != 0 or j != 0)
```
WDemidenko, line 101: Maybe usual loop would be better here, probablly will make it much easier to read

app/main.py
```diff
@@ -1,34 +1,137 @@
...
+
+    @staticmethod
+    def _create_ships(
+        ships: list[tuple[tuple[int, int], tuple[int, int]]]
+    ) -> dict[tuple[Any, ...], Ship]:
+        field = dict()
+        ships = [Ship(*coords) for coords in ships]
+        for ship in ships:
+            for deck in ship.decks:
+                field[tuple([deck.row, deck.column])] = ship
```
WDemidenko, line 122: Again, 2 times "for * in ships" cycle

WDemidenko (changes requested):

app/main.py
NEkropo1: not on my desk )

app/main.py
NEkropo1: They're totally different ) And they can be done only one after another, we can't combine them in one loop, only make different if's or asserts, to call 1 by 1 which is dumb, since you can cover all them in one comprehension + loop, which makes it more DRY and simple. And complexity of those loops is ~O(1))), look:
The method starts by creating a ships list by calling the values() method on the field dictionary. This has a time complexity of O(n), where n is the number of elements in the dictionary. We can't skip that, cause we're validating that data.

Next, the method creates the expected_counts and counts dictionaries. Since these dictionaries have at most 5 elements, their creation has a constant time complexity, i.e., O(1).

The method then enters a loop that iterates over each possible ship size, i.e., 0, 1, 2, 3, and 4. Inside the loop, it checks whether the number of ships of the current size in the ships list matches the expected count for that size. This operation has a time complexity of O(n), since it requires iterating over the ships list. Which is made from counting ships, so we would loop over them in all variations.

Last, the method enters another loop that iterates over each ship in the ships list. Inside the loop, it iterates over each deck in the ship and checks whether any of the deck's neighbors is also occupied by a ship. And this operation has a time complexity of O(n^2), since it requires iterating over each deck of each ship and then iterating over the neighboring cells of each deck. And I have no idea, how to fasten that.

app/main.py
NEkropo1: Well, I see this is efficient and easy to read, idk what you're talking about. If mentor will agree with you, I'll refactor that)

app/main.py
NEkropo1: list comprehension works much faster than creating objects in a loop and iterating through deck in them, this is because creating objects can be expensive and doing it multiple times inside a loop can slow down the program significantly(implementation issue).
I am creating all ships first and then looping through them to add each deck to a field dictionary. Guess it's a good approach in terms of performance as I'm only creating the ships once and then reusing them multiple times.
Therefore, I'm certain that creating ships before looping through them is generally the better approach in terms of performance and code clarity comparing to make them in a loop and working further.

I mean I could do something like this;
```
    field = dict()
    for coords in ships:
        ship = Ship(*coords)
        for deck in ship.decks:
            field[tuple([deck.row, deck.column])] = ship
    return field
```
But it won't be any faster, only would save a little bit of memory. Check this out:

image

WDemidenko (approved):
Anyway, I'm making suggestions, it's up to you whether you implement them or not

DmytryTong (approved):
Good job)

YuriiKindrat (approved):

PR #150, opened 2023-02-24
Solution

AlimFreight:


app/main.py
```diff
@@ -1,34 +1,106 @@
...
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+        if deck is not None:
+            deck.is_alive = False
+            self.is_drowned = all(
+                not deck.is_alive for deck in self.decks
+            )
+            return "Sunk!" if self.is_drowned else "Hit!"
+        else:
+            return "Miss!"
```
```suggestion
        return "Miss!"
```

app/main.py
```diff
@@ -1,34 +1,106 @@
...
+                            "Ships cannot be located in neighboring cells"
+                        )
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field.keys():
+            self.field[location].fire(location[0], location[1])
+            if self.field[location].is_drowned:
+                return "Sunk!"
+            return "Hit!"
+        return "Miss!"
```
Grayder0152, line 103: If i am not mistaken you can do that
```suggestion
        return self.field[location].fire(location[0], location[1])
```

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,34 +1,103 @@
...
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
```
ArsenPidhoretskyi: tuple of what? 
Add inner type

app/main.py
```diff
@@ -1,34 +1,103 @@
...
+    ) -> None:
+        self.decks = []
+        row1, col1 = start
+        row2, col2 = end
+        if row1 == row2:
+            for col in range(col1, col2 + 1):
+                self.decks.append(Deck(row1, col))
+        else:
+            for row in range(row1, row2 + 1):
+                self.decks.append(Deck(row, col1))
```
ArsenPidhoretskyi: You can do just in two loops:
```
for row in range(row1, row2 + 1):
    for col in range(col1, col2 + 1):
        self.decks.append(Deck(row, col))
```

app/main.py
```diff
@@ -1,34 +1,103 @@
...
-        # which is located in it
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
```
ArsenPidhoretskyi: ships list of what?

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,100 @@
...
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple[int],
```
ArsenPidhoretskyi: wrong type
must be tuple[int, int]
the same in other places)
tuple accept all arguments that inside

app/main.py
```diff
@@ -1,34 +1,100 @@
...
+        row2, col2 = end
+        for row in range(row1, row2 + 1):
+            for col in range(col1, col2 + 1):
+                self.decks.append(Deck(row, col))
+        self.is_drowned = is_drowned
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #153, opened 2023-02-25
Solution

WDemidenko:


app/main.py
```diff
@@ -1,34 +1,116 @@
...
+    def __init__(self, ships: list) -> None:
+        field = {}
+        for coord in ships:
+            ship = Ship(*coord)
+            for deck in ship.decks:
+                field[(deck.row, deck.column)] = ship
+        self.field = field
+
+    def _validate_field(self) -> str:
+        ships = list({ship for ship in self.field.values()})
```
amber-marichi: i dont think you need a comprehension structure here since .values() method already returns a list. you can just make set of it with no iterations :)

app/main.py
```diff
@@ -1,34 +1,116 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        field = {}
+        for coord in ships:
+            ship = Ship(*coord)
+            for deck in ship.decks:
+                field[(deck.row, deck.column)] = ship
```
amber-marichi: i guess it would be better to remove this logic out of __init__ . something like your set_up method from the Ship class is nicer imo. Also that way you could've done your validate_field checks after setting things up (if you need it done automatically of course)

amber-marichi (approved):
overall pretty cool. few suggestions from me

app/main.py
WDemidenko: You are absolutly right

app/main.py
```diff
@@ -1,34 +1,121 @@
...
+            ship = Ship(*coord)
+            for deck in ship.decks:
+                fields[(deck.row, deck.column)] = ship
+        ships = list(set(fields.values()))
+        sizes = []
+        for i, uniq_ship in enumerate(ships):
+            sizes.append(len(uniq_ship.decks))
+            for deck in uniq_ship.decks:
+                x, y = deck.row, deck.column
+                tests = [
```
wQuelS: tests? 
And also it can be moved somewhere else not to overload this function. It is already rather complicated

wQuelS (changes requested):

YuriiKindrat (approved):

Y-Havryliv (approved):

PR #164, opened 2023-03-02
Solution

yhwh6:


app/main.py
```diff
@@ -1,34 +1,68 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
Grayder0152: What if `get_deck` return None?

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
```
ArsenPidhoretskyi: Provide types of tuple items

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi: Example: tuple[str, str, int] - mean 1 parameter is str, second str, and third int

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        if self.start[0] == self.end[0]:
+            self.decks = [
+                Deck(self.start[0], i)
+                for i in range(self.start[1], self.end[1] + 1)
+            ]
+        if self.start[1] == self.end[1]:
+            self.decks = [
+                Deck(i, self.start[1])
+                for i in range(self.start[0], self.end[0] + 1)
+            ]
```
ArsenPidhoretskyi: You can simplify and use loop inside loop

app/main.py
```diff
@@ -1,34 +1,73 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+
+        if deck is None:
+            print(f"No deck found at row {row}, column {column}")
+            return
+
+        deck.is_alive = False
+        drowned = []
+
+        for deck in self.decks:
+            drowned.append(deck.is_alive)
```
ArsenPidhoretskyi: Remove this
You can just do:
if not any(deck.is_alive for deck in self.decks)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -19,18 +22,13 @@ def __init__(
...
-        if self.start[1] == self.end[1]:
-            self.decks = [
-                Deck(i, self.start[1])
-                for i in range(self.start[0], self.end[0] + 1)
-            ]
-
-    def get_deck(self, row: int, column: int) -> Deck:
+        self.decks = []
+
+        for st in range(self.start[0], self.end[0] + 1):
```
vasylhnatiuk: `st` and `en` are bad names for variables

vasylhnatiuk (changes requested):

vsmutok (approved):
Good job 👍

PR #168, opened 2023-03-09
Solution

StreetNik:


app/main.py
```diff
@@ -1,34 +1,117 @@
...
+
+        # list with count of different ships,
+        # different_ships_count[index + 1] = count of index-deck ships
+        different_ships_count = [0] * 4
+
+        for ship in ships:
+            different_ships_count[len(ship.decks) - 1] += 1
+
+        assert len(ships) == 10, \
+            ValueError("the total number of the ships should be 10")
```
ArsenPidhoretskyi: It is not correct, in this case better to use if len(ships)  == 10: and then raise ValueError

app/main.py
```diff
@@ -1,34 +1,117 @@
...
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
+    def __init__(
+            self,
+            start: list,
+            end: list,
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,123 @@
...
         # Create decks and save them to a list `self.decks`
-        pass
+        self.is_drowned = is_drowned
+        self.decks = []
+        if start[0] != end[0]:
+            for i in range(start[0], end[0] + 1):
+                self.decks.append(Deck(i, start[1]))
+        else:
+            for i in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], i))
```
katevasilenko: Does than make sense to move this part of code to separate method?

app/main.py
```diff
@@ -1,34 +1,123 @@
...
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
+    def __init__(
+            self,
+            start: Tuple[int, int],
+            end: Tuple[int, int],
+            is_drowned: bool = False
+    ) -> None:
         # Create decks and save them to a list `self.decks`
```
katevasilenko: Please remove comments of transform them to doc-strings all over the project

app/main.py
```diff
@@ -1,34 +1,123 @@
...
 
-    def fire(self, location: tuple):
+    def fire(self, location: Tuple[int, int]) -> str:
         # This function should check whether the location
         # is a key in the `self.field`
         # If it is, then it should check if this cell is the last alive
         # in the ship or not.
-        pass
+        if location in self.field:
+            res = self.field[location].fire(*location)
```
katevasilenko: `res` Is not very good name for variable

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+                deck_list.append(Deck(i, start[1]))
+        else:
+            for i in range(start[1], end[1] + 1):
+                deck_list.append(Deck(start[0], i))
+        return deck_list
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
Grayder0152: This function can return `None` .... Change typing

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
```
Grayder0152: what if desk is None?

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+
+        self._validate()
+
+    def fire(self, location: Tuple[int, int]) -> str:
+        if location in self.field:
+            result_string = self.field[location].fire(*location)
+            self.print_field()
+            return result_string
+        else:
+            return "Miss!"
```
```suggestion
        return "Miss!"
```

app/main.py
```diff
@@ -1,34 +1,119 @@
...
+
+        for index in range(4):
+            if different_ships_count[index] == 4 - index:
+                ValueError(f"there should be {4 - index} "
+                           f"{index + 1}-deck ships")
+
+        for cell in self.field:
+            for other_cell in self.field:
+                if self.field[cell] != self.field[other_cell]:
+                    if abs(cell[0] - other_cell[0]) < 2 \
```
Grayder0152: use `()` instead of `\`

Grayder0152 (changes requested):

vsmutok (approved):
Good Job 👍

PR #173, opened 2023-03-18
Add solution

ZaikaDaria:


app/main.py
```diff
@@ -1,34 +1,63 @@
...
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple[tuple]]) -> None:
+        self.fields = {}
+        for ship in ships:
+            self.ship = Ship(ship[0], ship[1])
```
katevasilenko:
```suggestion
            ship = Ship(ship[0], ship[1])
```
I believe you don’t need ship as an attribute of instance

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,34 +1,63 @@
...
-        pass
+        for row in range(start[0], end[0] + 1):
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row, column))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
Grayder0152, line 33:
```suggestion
    def get_deck(self, row: int, column: int) -> Oprional[Deck]:
```

Grayder0152 (approved):

PR #174, opened 2023-03-19
Solution

Roman28101:


app/main.py
```diff
@@ -1,34 +1,56 @@
...
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self, start: tuple[int, int],
```
katevasilenko, line 30: Please put each parameter to a new line
```suggestion
            self, 
            start: tuple[int, int],
```

app/main.py
```diff
@@ -1,34 +1,56 @@
...
 
 class Battleship:
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.ships = ships
```
katevasilenko: I believe you don’t need ship as an attribute of instance here

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,34 +1,56 @@
...
+            end: tuple[int, int],
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        for row in range(self.start[0], self.end[0] + 1):
+            for column in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(row, column))
```
vsmutok: put it in a separate method

vsmutok (changes requested):

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+            row: int,
+            column: int,
+            is_alive: bool = True
+    ) -> None:
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
+
+    @staticmethod
+    def decks_group(start: tuple[int, int], end: tuple[int, int]) -> list:
```

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = Deck.decks_group(start, end)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

ArsenPidhoretskyi: Because if in the loop we do not find any deck we return None as default

app/main.py
```diff
@@ -1,34 +1,61 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        for deck in self.decks:
+            self.is_drowned = True
+            if deck.is_alive:
+                self.is_drowned = False
```
ArsenPidhoretskyi: You can use 
self.is_drowned = all( and here some loop)

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #177, opened 2023-03-25
Solution

Bohdan2001007:


app/main.py
```diff
@@ -1,34 +1,57 @@
...
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: Tuple[int],
```
ArsenPidhoretskyi: Tuple with one parameter?)
```suggestion
            start: Tuple[int, int],
```
Change in all places

app/main.py
```diff
@@ -1,34 +1,57 @@
...
+        self.decks = [
+            Deck(row, column)
+            for column in range(start[1], end[1] + 1)
+            for row in range(start[0], end[0] + 1)
+        ]
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Union[Deck, None]:
```
```suggestion
    def get_deck(self, row: int, column: int) -> Optional[Deck]:
```

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #179, opened 2023-03-27
Solution

yuraua5:
Unfortunately, I don't know how to do the last additional task: ships shouldn't be located in the neighboring cells (even if cells are neighbors by diagonal).

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,34 +1,103 @@
+import numpy
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(self,
+                 row: int,
+                 column: int,
+                 is_alive: bool = True,
+                 deck_print: str = u"\u25A1") -> None:
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
+        self.deck_print = deck_print
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self,
+                 start: tuple,
+                 end: tuple,
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,34 +1,103 @@
...
+        self.is_drowned = is_drowned
+        self.decks = []
+        if self.start[0] == self.end[0]:
+            for i in range(self.start[1], self.end[1] + 1):
+                self.decks.append(Deck(self.start[0], i))
+            return
+        if self.start[1] == self.end[1]:
+            for i in range(self.start[0], self.end[0] + 1):
+                self.decks.append(Deck(i, self.start[1]))
+            return
```
ArsenPidhoretskyi: Use loop in loop

app/main.py
```diff
@@ -1,34 +1,103 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,103 @@
...
+        self.ships = ships
+        self.field = {}
+        for ship in self.ships:
+            self.field[ship] = Ship(ship[0], ship[1])
+        self._validate_field(self.field)
+
+    def fire(self, location: tuple) -> str:
+        for key, value in self.field.items():
+            if (location[0] in range(key[0][0], key[1][0] + 1)
+                    and location[1] in range(key[0][1], key[1][1] + 1)):
```
ArsenPidhoretskyi: So hard logic/ Please simplify this
So much indexes

ArsenPidhoretskyi: Key, value also not clear

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
You committed and pushed nothing

vasylhnatiuk (approved):
Good job!

PR #190, opened 2023-04-20
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1,34 +1,61 @@
...
+                 start: tuple,
+                 end: tuple,
+                 is_drowned: bool = False) -> None:
+        self.decks = [Deck(start[0], start[1])]
+        step = end[0] - start[0] + end[1] - start[1]
+        for i in range(1, step + 1):
+            if end[0] == start[0]:
+                self.decks.append(Deck(start[0], start[1] + i))
+            else:
+                self.decks.append(Deck(start[0] + i, start[1]))
```
ArsenPidhoretskyi: Simplify and just use double loop because it so hard to understand 
```
for row_index in range:
    for column_index in range:
```

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+        self.is_drowned = is_drowned
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self,
+                 row: int,
+                 column: int) -> Deck:
```
ArsenPidhoretskyi: Can return None also

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+        self.field = {}
+        for ship in ships:
+            boat = Ship(ship[0], ship[1])
+            self.field[ship[0]] = boat
+            step = ship[1][0] - ship[0][0] + ship[1][1] - ship[0][1]
+            for i in range(1, step + 1):
+                if ship[0][0] == ship[1][0]:
+                    self.field[(ship[0][0], ship[0][1] + i)] = boat
+                else:
+                    self.field[(ship[0][0] + i, ship[0][1])] = boat
```
ArsenPidhoretskyi: Just iterate through decks)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,53 @@
...
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self,
+             row: int,
+             column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
ArsenPidhoretskyi: self.get_deck(row, column)
can return None
First of all you must check that desk is not None

app/main.py
```diff
@@ -1,34 +1,53 @@
...
+                self.field[(deck.row, deck.column)] = my_ship
 
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def fire(self,
+             location: tuple) -> str:
```
```suggestion
    def fire(self, location: tuple) -> str:
```

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @HalynaPetrova not fixed

app/main.py
```diff
@@ -1,34 +1,48 @@
...
+            for column_index in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row_index, column_index))
+        self.is_drowned = is_drowned
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck is not None:
```
ArsenPidhoretskyi: this check is not needed

app/main.py
ArsenPidhoretskyi: You must check here if self.get_deck() return None

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #191, opened 2023-04-23
Solution

goldenuni:


app/main.py
```diff
@@ -1,34 +1,92 @@
...
+                 start: tuple[int],
+                 end: tuple[int],
+                 is_drowned: bool = False) -> None:
+        self.decks = [Deck(start[0], start[1])]
+        if start[0] == end[0]:
+            for i in range(start[1] + 1, end[1] + 1):
+                self.decks.append(Deck(start[0], i))
+        else:
+            for i in range(start[0] + 1, end[0] + 1):
+                self.decks.append(Deck(i, start[1]))
```
ArsenPidhoretskyi, line 22: You can just use loop inside loop instead of if statement

ArsenPidhoretskyi (approved):
Good job!

PR #195, opened 2023-04-29
solution

OleksandrGrk:


app/main.py
```diff
@@ -1,34 +1,92 @@
...
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.decks = []
+        if start[0] == end[0]:
+            for i in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], i))
+        else:
+            for i in range(start[0], end[0] + 1):
+                self.decks.append(Deck(i, end[1]))
```
ArsenPidhoretskyi: You can do it in the double loop instead of if-else

app/main.py
```diff
@@ -1,34 +1,92 @@
...
+        self.is_drowned = is_drowned
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> bool:
```
ArsenPidhoretskyi: get_deck due to name must return deck not bool

app/main.py
```diff
@@ -1,34 +1,92 @@
...
+        for deck in self.decks:
+            return (row, column) in deck.cell
+
+    def fire(self, row: int, column: int) -> None:
+        for deck in self.decks:
+            print(deck.cell)
```
ArsenPidhoretskyi: Remove this code

app/main.py
```diff
@@ -1,34 +1,92 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        self.battle_field = []
+        for _ in range(10):
+            self.battle_field.append(["~" for _ in range(10)])
```
ArsenPidhoretskyi: It is redundant and better to do it in the loop due to the current ships and decks
Please Simplify code

ArsenPidhoretskyi (changes requested):

app/main.py
OleksandrGrk: When I use double for loop I get duplicate of identical decks. It ruins code.

OleksandrGrk: />
```python

```  _ self.decks = []
>         for i in range(start[1], end[1] + 1):
>             self.decks.append(Deck(start[0], i))
>         for e in range(start[0], end[0] + 1):
>             self.decks.append(Deck(e, end[1]))
>         self.is_drowned = `is_drowned_`

File: app/main.py (hunk 4 above)
Comment (OleksandrGrk) [reply app/main.py]: Didn't catch the idea.. Where do I need to create battle field? In Ship class?

File: app/main.py (hunk 2 above)
Comment (Grayder0152) [reply app/main.py]: pls resolve comment

**Review by Grayder0152 — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 1 above)
Comment (ArsenPidhoretskyi) [reply app/main.py]:
```
for row_index in range(start[0], end[0] + 1):
    for column_index in range(start[1], end[1] + 1):
        self.decks.append(Deck(row_index, column_index))
```

File: app/main.py (hunk 4 above)
Comment (ArsenPidhoretskyi) [reply app/main.py]: In `__str__` method you can just print due to current ships/decks no need to save battle_field inside class

File: app/main.py [hunk 5]
```diff
@@ -1,34 +1,90 @@
...
+
+        for ship in ships:
+            self.field[ship] = Ship(ship[0], ship[1])
```
ArsenPidhoretskyi: Better to save inside field coordinates of each deck which will equal to ship
And than you can simplify fire method

ArsenPidhoretskyi (changes requested):

app/main.py
OleksandrGrk: In which do I need to implement this method? 
I write new print func inside Battleship class. Did I write it correctly? Do I need to move this code to magic _ _str_ _ method?

app/main.py
```diff
@@ -1,34 +1,93 @@
...
+
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for _ship in ships:
+            self.field[_ship] = Ship(_ship[0], _ship[1])
+
+        self._validate_field()
+
+    def fire(self, location: tuple) -> str:
+        for _ship in self.field.values():
```
```suggestion
        for ship in self.field.values():
```

app/main.py
```diff
@@ -1,34 +1,93 @@
...
+            for _deck in _ship.decks:
```
```suggestion
            for deck in _ship.decks:
```

app/main.py
```diff
@@ -1,34 +1,93 @@
...
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for _ship in ships:
+            self.field[_ship] = Ship(_ship[0], _ship[1])
```
```suggestion
            self.field[ship] = Ship(ship[0], ship[1])
```
Why you need these in all places???

app/main.py
```diff
@@ -1,34 +1,93 @@
...
+        self._validate_field()
+
+    def fire(self, location: tuple) -> str:
+        for _ship in self.field.values():
+            for _deck in _ship.decks:
+                if _deck.cell == location:
+                    _deck.is_alive = False
+                    are_any_alive = [_deck.is_alive for _deck in _ship.decks]
+                    if not any(are_any_alive):
+                        _ship.is_drowned = True
```
ArsenPidhoretskyi: Set this inside class ship. It related to ship, not to Battlefield

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,97 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        self.decks = []
+        for row_index in range(start[0], end[0] + 1):
+            for column_index in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row_index, column_index))
+        self.is_drowned = is_drowned
+
+    def get_deck(self, row: int, column: int) -> tuple:
```
ArsenPidhoretskyi: No get_deck must return Optional[Deck]

app/main.py
```diff
@@ -1,34 +1,97 @@
...
+        for deck in self.decks:
+            if (row, column) in deck.cell:
```
```suggestion
            if (row, column) == deck.cell:
```

app/main.py
```diff
@@ -1,34 +1,97 @@
...
+                return deck.cell
```
ArsenPidhoretskyi: Return deck in this case

app/main.py
```diff
@@ -1,34 +1,97 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        for deck in self.decks:
+            if (row, column) == deck.cell:
```
ArsenPidhoretskyi: That use deck = self.get_deck(row, column)

app/main.py
```diff
@@ -1,34 +1,97 @@
...
+        if all((
+                len_is_10,
+                one_four_deck,
+                two_triple_deck,
+                three_double_deck,
+                four_single_deck
+        )):
+            print("Field is validated")
+        else:
+            print("You break the rules of battleship game")
```
ArsenPidhoretskyi: Better to raise an Exception is something went wrong

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,97 @@
...
-        # And update the `is_drowned` value if it's needed
-        pass
+        self.decks = []
+        for row_index in range(start[0], end[0] + 1):
+            for column_index in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row_index, column_index))
+        self.is_drowned = is_drowned
+        print("decks: ", self.decks)
+
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: It return Optional[Deck]

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #204, opened 2023-05-18
solution

mblazhko:


app/main.py
```diff
@@ -1,34 +1,51 @@
...
 
-    def fire(self, row, column):
+    def fire(self, row: int, column: int) -> str:
         # Change the `is_alive` status of the deck
         # And update the `is_drowned` value if it's needed
-        pass
+        for deck in self.decks:
+            if self.get_deck(row, column) == deck:
+                deck.is_alive = False
+                if any([decks.is_alive for decks in self.decks]) is False:
```
olena-hul:
```suggestion
                if any(decks.is_alive for decks in self.decks) is False:
```

app/main.py
```diff
@@ -1,34 +1,51 @@
...
-        # which is located in it
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.ships = [Ship(deck[0], deck[1]) for deck in ships]
+        self.field = {}
+        for ship in self.ships:
+            for row in range(10):
+                for column in range(10):
+                    if ship.get_deck(row, column):
+                        self.field[(row, column)] = ship
```
olena-hul: You can create a ship inside for loop and then loop through decks and assign them to the field - it will be more optimized

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #207, opened 2023-05-19
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1,34 +1,65 @@
...
+        elif start[0] == end[0]:
+            current_column = start[1]
+            while current_column != end[1] + 1:
+                self.decks.append(Deck(start[0], current_column))
+                current_column += 1
+        elif start[1] == end[1]:
+            current_row = start[0]
+            while current_row != end[0] + 1:
+                self.decks.append(Deck(current_row, start[1]))
+                current_row += 1
```
ArsenPidhoretskyi: So hard code. Use loop inside loop instead of checks

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+        print(self.decks)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Union[Deck | None]:
```
```suggestion
    def get_deck(self, row: int, column: int) -> Optional[Deck]:
```

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> str:
+        current_deck = self.get_deck(row, column)
+        if current_deck.is_alive:
+            current_deck.is_alive = False
+            if not self.is_drowned:
+                self.is_drowned = True
+            if not any(deck.is_alive for deck in self.decks):
```
```suggestion
            if self.is_drowned:
```
Keep your code simple

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> str:
+        current_deck = self.get_deck(row, column)
+        if current_deck.is_alive:
+            current_deck.is_alive = False
+            if not self.is_drowned:
+                self.is_drowned = True
```
ArsenPidhoretskyi: This check is not true
```suggestion
           self.is_drowned = not any(deck.is_alive for deck in self.decks)
```

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+        self.field = {}
+        for ship in ships:
+            ship_instance = Ship(ship[0], ship[1])
+            for deck in ship_instance.decks:
+                self.field[(deck.row, deck.column)] = ship_instance
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field.keys():
+            return self.field[location].fire(location[0], location[1])
+        else:
```
ArsenPidhoretskyi: else: is not needed

ArsenPidhoretskyi (changes requested):

app/main.py
V-Shkrobatskyi: but then when code will return "Miss!"?

app/main.py
V-Shkrobatskyi: can't imagine it... i will try to use just two loops

app/main.py
V-Shkrobatskyi: oh, I understand, you mean just delete "else:" without code below

app/main.py
```diff
@@ -1,34 +1,56 @@
...
+            if (row, column) == (deck.row, deck.column):
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> str:
+        current_deck = self.get_deck(row, column)
+        if current_deck.is_alive:
```
```suggestion
        if current_deck is not None and current_deck.is_alive:
```

Grayder0152 (approved):

app/main.py
V-Shkrobatskyi: after all actions in class Battleship it looks like impossible to get None from current_deck()

PR #210, opened 2023-05-23
Solution

OlehOryshchuk:


Bohdan-Salamakha (changes requested):
single letter can't be a variable name

app/main.py
```diff
@@ -1,34 +1,139 @@
...
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(self,
+                 row: int,
+                 column: int,
+                 is_alive: bool = True
+                 ) -> None:
```
Bohdan-Salamakha:
```suggestion
    def __init__(
                 self,
                 row: int,
                 column: int,
                 is_alive: bool = True
                 ) -> None:
```

app/main.py
```diff
@@ -1,34 +1,139 @@
...
-        pass
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def __init__(self,
+                 start: Tuple[int],
+                 end: Tuple[int],
+                 is_drowned: bool = False
+                 ) -> None:
```
Bohdan-Salamakha:
```suggestion
    def __init__(
                 self,
                 start: Tuple[int],
                 end: Tuple[int],
                 is_drowned: bool = False
                 ) -> None:
```

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #211, opened 2023-05-25
:' ) without interface

nazarbychkar:


app/main.py
```diff
@@ -1,34 +1,89 @@
...
+            if x_start - x_end < 0:
+                self.decks = [
+                    Deck(x, y_start)
+                    for x in range(x_start, x_end + 1)
+                ]
+            else:
+                self.decks = [
+                    Deck(x, y_start)
+                    for x in range(x_end, x_start + 1)
+                ]
```
ArsenPidhoretskyi: Simplify this part of code and just use loop inside loop:
```
for row_index in range(x_start, x_end):
     for column_index in range(y_start, y_end):
```

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: None also possible to return

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+                return deck
 
-    def fire(self, row, column):
+    def fire(self, row: int, column: int) -> None:
         # Change the `is_alive` status of the deck
         # And update the `is_drowned` value if it's needed
-        pass
+        self.get_deck(row, column).is_alive = False
+        if all(deck.is_alive is False for deck in self.decks):
+            self.is_drowned = True
```
```suggestion
        self.is_drowned = all(deck.is_alive is False for deck in self.decks):
```

app/main.py
```diff
@@ -1,34 +1,89 @@
...
         # If it is, then it should check if this cell is the last alive
         # in the ship or not.
-        pass
+        if self.field.get(location):
+            ship = self.field.get(location)
+            live_deck = ship.get_deck(*location).is_alive is True
+            last_live_deck = [
+                deck.is_alive
+                for deck in ship.decks
+            ].count(True) == 1
```
ArsenPidhoretskyi: This is redundant code. You can just return result depended on ship.is_drowned

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+        for deck in self.decks:
+            if row == deck.row and column == deck.column:
+                return deck
 
-    def fire(self, row, column):
+    def fire(self, row: int, column: int) -> None:
         # Change the `is_alive` status of the deck
         # And update the `is_drowned` value if it's needed
-        pass
+        self.get_deck(row, column).is_alive = False
```
ArsenPidhoretskyi: check that self.get_deck(row, column) returns not None before to set = False
```suggestion
        deck = self.get_deck(row, column)
        if deck is not None:
             deck.is_alive = False
```

app/main.py
```diff
@@ -1,34 +1,65 @@
...
 
-    def fire(self, location: tuple):
+    def fire(self, location: tuple) -> str:
         # This function should check whether the location
         # is a key in the `self.field`
         # If it is, then it should check if this cell is the last alive
         # in the ship or not.
-        pass
+        if self.field.get(location):
+            ship = self.field.get(location)
```
ArsenPidhoretskyi: Write code clear as possible
```suggestion
        ship = self.field.get(location)
        if ship is not None:
```

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+            live_deck = ship.get_deck(*location).is_alive
+            ship.fire(*location)
+            if live_deck and ship.is_drowned:
+                return "Sunk!"
+            else:
+                ship.fire(*location)
```
ArsenPidhoretskyi: This line is strance. Because you already do file

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+                return "Hit!"
```
```suggestion
             return "Hit!"
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,67 @@
...
+            if row == deck.row and column == deck.column:
+                return deck
 
-    def fire(self, row, column):
+    def fire(self, row: int, column: int) -> None:
         # Change the `is_alive` status of the deck
         # And update the `is_drowned` value if it's needed
-        pass
+        deck = self.get_deck(row, column)
+        if deck:
```
```suggestion
        if deck is not None:
```

app/main.py
```diff
@@ -1,34 +1,67 @@
...
 
-    def fire(self, location: tuple):
+    def fire(self, location: tuple) -> str:
         # This function should check whether the location
         # is a key in the `self.field`
         # If it is, then it should check if this cell is the last alive
         # in the ship or not.
-        pass
+        ship = self.field.get(location)
+        if ship:
```
```suggestion
        if ship is not None:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #216, opened 2023-05-30
'Solution'

KirillMelanich:


app/main.py
```diff
@@ -1,34 +1,57 @@
...
-        pass
+        for row in range(start[0], end[0] + 1):
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row, column))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
```suggestion
    def get_deck(self, row: int, column: int) -> Deck | None:
```

app/main.py
```diff
@@ -1,34 +1,57 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
Grayder0152: What if `get_deck` return None?

app/main.py
```diff
@@ -1,34 +1,57 @@
...
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self,
+                 ships: List[Tuple[Tuple[int, int], Tuple[int, int]]]
```
Grayder0152: or you can use built-in types
```suggestion
                 ships: list[tuple[tuple[int, int], tuple[int, int]]]
```

app/main.py
```diff
@@ -1,34 +1,57 @@
+from typing import Dict, Tuple, List
```
Grayder0152: try instead of this types use built-ins types

app/main.py
```diff
@@ -1,34 +1,57 @@
...
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    battle_array: List[List[str]] = []
```
Grayder0152: Why do you use it variable ?

Grayder0152 (changes requested):

Grayder0152 (changes requested):
Pls, resolve comments

app/main.py
```diff
@@ -1,34 +1,61 @@
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    battle_array: list[list[str]] = []
```
Grayder0152, line 2: You can  remove it, if not use

Grayder0152 (approved):

PR #217, opened 2023-05-31
Solution

markus2812:


app/main.py
```diff
@@ -1,34 +1,65 @@
...
+        decks = []
+        if self.start[0] == self.end[0]:
+            for column in range(self.start[1], self.end[1] + 1):
+                deck = Deck(self.start[0], column)
+                decks.append(deck)
+        else:
+            for row in range(self.start[0], self.end[0] + 1):
+                deck = Deck(row, self.start[1])
+                decks.append(deck)
+        return decks
```
vsmutok: try to improve it

hint 
you can use `sorted()` and list comprehensin

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        pass
+    def __init__(self,
+                 ships: List[Tuple[Tuple[int, int], Tuple[int, int]]])\
+            -> None:
+        self.field = {}
+        for ship in ships:
+            ship_obj = Ship(ship[0], ship[1])
+            for row in range(ship[0][0], ship[1][0] + 1):
+                for column in range(ship[0][1], ship[1][1] + 1):
+                    self.field[(row, column)] = ship_obj
```
vsmutok: try to optimize it and create a shipі using list comprehension

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self,
+                 ships: List[Tuple[Tuple[int, int], Tuple[int, int]]])\
+            -> None:
+        self.field = {}
+        for ship in ships:
+            ship_obj = Ship(ship[0], ship[1])
```
```suggestion
            ship_obj = Ship(*ship)
```

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+            for row in range(ship[0][0], ship[1][0] + 1):
+                for column in range(ship[0][1], ship[1][1] + 1):
+                    self.field[(row, column)] = ship_obj
+
+    def fire(self, location: Tuple[int, int]) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            return ship.fire(location[0], location[1])
```
```suggestion
            return ship.fire(*location)
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,34 +1,62 @@
...
-        # Find the corresponding deck in the list
-        pass
+    def create_decks(self) -> List[Deck]:
+        if self.start[0] == self.end[0]:
+            return ([Deck(self.start[0],
+                          column) for column in
+                     range(self.start[1], self.end[1] + 1)])
+        else:
+            return ([Deck(row, self.start[1]) for row in
+                     range(self.start[0], self.end[0] + 1)])
```
ArsenPidhoretskyi: Use loop inside loop instead of if-else:
```
return [
    Deck(row, column)
    for row in range(self.start[0], self.end[0] + 1)]:
    for column in range(self.start[1], self.end[1] + 1)]
]
```

app/main.py
```diff
@@ -1,34 +1,62 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

ArsenPidhoretskyi: If deck will be not found ^^

app/main.py
```diff
@@ -1,34 +1,62 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            if all(not d.is_alive for d in self.decks):
```
ArsenPidhoretskyi: rename d to deck

app/main.py
```diff
@@ -1,34 +1,62 @@
...
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self,
+                 ships: List[Tuple[Tuple[int, int], Tuple[int, int]]])\
+            -> None:
```
```suggestion
    def __init__(
        self, ships: List[Tuple[Tuple[int, int], Tuple[int, int]]]
    ) -> None:
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #219, opened 2023-06-01
'Solution#BattleShip'

Jhaivir:


app/main.py
```diff
@@ -1,34 +1,173 @@
...
+        if end[0] == start[0]:
+            self.decks = [
+                Deck(start[0], start[1] + i)
+                for i in range(end[1] - start[1] + 1)
+            ]
+        else:
+            self.decks = [
+                Deck(start[0] + i, start[1])
+                for i in range(end[0] - start[0] + 1)
+            ]
```
ArsenPidhoretskyi: You can use loop inside loop

app/main.py
```diff
@@ -1,34 +1,173 @@
...
+        self.is_drowned = is_drowned
         # Create decks and save them to a list `self.decks`
-        pass
 
-    def get_deck(self, row, column):
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,173 @@
...
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+        [" ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v", " ^v"],
+    ]
```
ArsenPidhoretskyi: First of all, it must be value of instance , not class

ArsenPidhoretskyi (changes requested):

app/main.py
Jhaivir: Not understood

vasylhnatiuk: try to do it like this 

`        self.decks = [
            Deck(row, column)
            for row in range(start[0], end[0] + 1)
            for column in range(start[1], end[1] + 1)
        ]`

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,162 @@
...
+    border_d,
+    border_l,
+    border_r,
+    border_u,
+    hit,
+    miss,
+    sunk,
+
+
+)
```
vsmutok: Maybe better like this 
```suggestion
from app.standard import *
```

app/main.py
```diff
@@ -1,34 +1,162 @@
...
+        for ship in ships:
+            new_ship = Ship(ship[0], ship[1])
+            self.ships.append(new_ship)
+            for deck in new_ship.decks:
+                self.art_field[deck.row][deck.column] = art_deck
+                self.field[(deck.row, deck.column)] = new_ship
+        self._validate_field(self.ships)
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field.keys():
```
vsmutok: `keys()` by default 
```suggestion
        if location in self.field:
```

app/main.py
```diff
@@ -1,34 +1,162 @@
...
+            return "Hit!"
+        self.art_field[location[0]][location[1]] = miss
+        return "Miss!"
+
+    def print_field(self) -> None:
+        print(border_u)
+        for row in self.art_field:
+            print(border_l + "".join(row) + border_r)
+        print(border_d)
+        print("\n")
```
```suggestion
        print(border_d + "\n")
```

vsmutok (changes requested):

app/main.py
Jhaivir: flake8 does not allow

vsmutok (approved):
Great

PR #220, opened 2023-06-02
'Solution'

Evgen-Shepelenko:


app/main.py
```diff
@@ -1,34 +1,91 @@
...
+        deck_ships = {4: 0, 3: 0, 2: 0, 1: 0}
+        for ship in ships:
+            deck_ships[len(ship.decks)] += 1
+            for row in range(ship.start[0] - 1, ship.start[0] + 2, 2):
+                for column in range(ship.start[1] - 1, ship.start[1] + 2, 2):
+                    if ((row, column) in self.field) \
+                            and (self.field[(row, column)] != ship):
+                        print("ships shouldn't be located "
+                              "in the neighboring cells")
+            for row in range(ship.end[0] - 1, ship.end[0] + 2, 2):
```
vasylhnatiuk: try to write additional function and to avoid duplicating loop

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+        self.is_drowned = False
+        self.decks = []
+        if start[0] == end[0] and start[1] == end[1]:
+            self.decks.append(Deck(start[0], end[1]))
+        elif start[0] == end[0]:
+            for column_number in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], column_number))
+        elif start[1] == end[1]:
+            for row_number in range(start[0], end[0] + 1):
+                self.decks.append(Deck(row_number, start[1]))
```
ArsenPidhoretskyi: Simplify this code and just use loop inside loop

app/main.py
```diff
@@ -1,34 +1,89 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: It also can return None

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+            if desk.row == row and desk.column == column:
+                return desk
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> str:
+        fire_desk = self.get_deck(row, column)
+        fire_desk.is_alive = False
```
ArsenPidhoretskyi, line 27: fire_deck can be None

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+        for desk in self.decks:
+            if desk.is_alive:
+                return "Hit!"
+        self.is_drowned = True
```
```suggestion
        self.is_drowned = not any(deck.is_alive for deck in self.decks)
        if self.is_drowned:
              return "Sunk!"
        return "Hit!"
```

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+                  "there should be 2 three-deck ships;\n"
+                  "there should be 1 four-deck ship;")
+
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for tuple_ship in ships:
+            ship = Ship(tuple_ship[0], tuple_ship[1])
+            if len(ship.decks) == 1:
+                self.field[tuple_ship[0]] = ship
+            else:
```
ArsenPidhoretskyi: This if is redundant code in else works will do the same

app/main.py
```diff
@@ -1,34 +1,89 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def check_neighbor_cells(self, ship: Ship, ship_desk: tuple) -> None:
+        for row in range(ship_desk[0] - 1, ship_desk[0] + 2, 2):
+            for column in range(ship_desk[1] - 1, ship_desk[1] + 2, 2):
+                if ((row, column) in self.field) \
+                        and (self.field[(row, column)] != ship):
+                    print("ships shouldn't be located "
```
ArsenPidhoretskyi: use raise exception instead of print

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+
+    def _validate_field(self) -> None:
+        ships = set([ship for ship in self.field.values()])
+        deck_ships = {4: 0, 3: 0, 2: 0, 1: 0}
+        for ship in ships:
+            deck_ships[len(ship.decks)] += 1
+            self.check_neighbor_cells(ship, ship.start)
+            self.check_neighbor_cells(ship, ship.end)
+        if deck_ships != {4: 1, 3: 2, 2: 3, 1: 4}:
+            print("the total number of the ships should be 10:\n"
```
ArsenPidhoretskyi: use raise exception instead of print

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #224, opened 2023-06-07
Solution

WalDragnir:


app/main.py
```diff
@@ -1,34 +1,150 @@
...
+        start_row, start_column = self.start
+        end_row, end_column = self.end
+        if start_row == end_row:
+            for column in range(start_column, end_column + 1):
+                self.decks.append(Deck(start_row, column))
+        elif start_column == end_column:
+            for row in range(start_row, end_row + 1):
+                self.decks.append(Deck(row, start_column))
+        else:
+            raise ValueError("Invalid ship position.")
```
ArsenPidhoretskyi: Use loop inside loop instead of if-else

app/main.py
```diff
@@ -1,34 +1,150 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
```
ArsenPidhoretskyi: It will do this auto. You can remove this line

app/main.py
```diff
@@ -1,34 +1,150 @@
...
+                or double_deck != 3
+                or three_deck != 2
+                or four_deck != 1
+        ):
+            raise ValueError("Invalid ship configuration.")
+
+        for ship1 in self._ships:
+            for ship2 in self._ships:
+                if ship1 is not ship2:
+                    if self._is_adjacent(ship1, ship2):
```
ArsenPidhoretskyi: Use and statement

app/main.py
```diff
@@ -1,34 +1,150 @@
...
+                for column in range(start[1], end[1] + 1):
+                    self.field[(row, column)] = new_ship
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            ship.fire(*location)
+            if ship.is_drowned:
+                return "Sunk!"
+            else:
```
ArsenPidhoretskyi: else: is not needed

app/main.py
```diff
@@ -1,34 +1,150 @@
...
+                return "Hit!"
+        else:
```
ArsenPidhoretskyi: else: is redudant

app/main.py
```diff
@@ -1,34 +1,150 @@
...
+            return "Miss!"
```
```suggestion
        return "Miss!"
```

app/main.py
```diff
@@ -1,34 +1,150 @@
...
+                    self.field[(row, column)] = new_ship
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            ship.fire(*location)
+            if ship.is_drowned:
+                return "Sunk!"
+            else:
+                return "Hit!"
```
```suggestion
            return "Hit!"
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #233, opened 2023-06-29
Solution

ShevchukVitalii:


app/main.py
```diff
@@ -1,34 +1,89 @@
...
+        decks = list()
+        if self.start[0] == self.end[0]:
+            for column in range(self.start[1], self.end[1] + 1):
+                deck = Deck(self.start[0], column)
+                decks.append(deck)
+        elif self.start[1] == self.end[1]:
+            for row in range(self.start[0], self.end[0] + 1):
+                deck = Deck(row, self.start[1])
+                decks.append(deck)
+        return decks
```
ArsenPidhoretskyi: Instead of if statements use loop inside loop

app/main.py
```diff
@@ -1,34 +1,89 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: It also can return None due to code

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            if self.is_drowned:
+                return "Hit!"
```
ArsenPidhoretskyi: Remove this if

ArsenPidhoretskyi (changes requested):

app/main.py
ShevchukVitalii: Fixed

ArsenPidhoretskyi (approved):
Well done!

PR #234, opened 2023-07-02
Solution

arturiermolenko:
In Extra tasks some validations decided to make before the field is created, as it is not necessary to create field with incorrect values of incoming data.


YukihiroSM (approved):
All looks good to me.

PR #235, opened 2023-07-03
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -1,34 +1,210 @@
...
+        for location in self.ships:
+            ships.append(Ship(start=location[0], end=location[1]))
+
+        for ship in ships:
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
+
+    def print_field(self) -> None:
+        whole_field = dict()
+        for i in range(10):
```
Dimosphen1: Avoid using `i` variable name, you could consider using `row_index`.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+            for index in range(10):
```
Dimosphen1: Consider renaming the `index` variable as well.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+                print("")
+            print(f"{whole_field.get(location)}      ", end="")
+
+    def _validate_field(self) -> bool:
+        single_deck_ship = 0
+        double_deck_ship = 0
+        three_deck_ship = 0
+        four_deck_ship = 0
+
+        ships = set([ship for ship in self.field.values()])
```
Dimosphen1: Do not use list comprehension here.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+            if ship_length == 3:
+                three_deck_ship += 1
+                continue
+            if ship_length == 4:
+                four_deck_ship += 1
+
+        if (
+                single_deck_ship == 4 and double_deck_ship == 3
+                and three_deck_ship == 2 and four_deck_ship == 1
+        ):
```
Dimosphen1: Instead of using three `and` statements consider using `all` here to check conditions.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+                single_deck_ship += 1
+                continue
+            if ship_length == 2:
+                double_deck_ship += 1
+                continue
+            if ship_length == 3:
+                three_deck_ship += 1
+                continue
+            if ship_length == 4:
+                four_deck_ship += 1
```
Dimosphen1: Consider using `dict` here to store ship length as the key and the number of ships with particular lengths as values. Increment `dict` values instead of 4 variables, it will avoid the use of 4 `if` conditions.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+        return True
+
+    def check_cell(
+            self,
+            locations: list[tuple],
+            location: tuple,
+            row: int,
+            column: int
+    ) -> bool:
+        if (location[0] + row, location[1] + column) in self.field:
```
Dimosphen1: Write the result of `(location[0] + row, location[1] + column)` to a separate variable and use it in checks.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+            if (location[0] + row, location[1] + column) not in locations:
+                return False
+        return True
```
Dimosphen1: Rewrite this method with the use of `all`, one return statement, and without `if` conditions.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+                if not self.check_cell(locations, location, 1, 1):
+                    return False
+                if not self.check_cell(locations, location, 0, 1):
+                    return False
+                if not self.check_cell(locations, location, -1, 1):
+                    return False
+                if not self.check_cell(locations, location, -1, 0):
+                    return False
+                if not self.check_cell(locations, location, -1, -1):
+                    return False
```
Dimosphen1: Consider using `itertools` module to generate necessary combinations and do not use 8 `if` statements here.

app/main.py
```diff
@@ -1,34 +1,210 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
+
+        for deck in self.decks:
+            if deck.is_alive:
+                self.is_drowned = False
+                break
+        else:
```
Dimosphen1, line 94: `else` indentation should be in line with `if`.

Dimosphen1 (changes requested):

app/main.py
NazarLytvyniuk: Thanks for your comments.
Here "else" belongs to the "for" statement

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+                continue
+            count_ships[ship_length] = 1
+
+        checks = (
+            count_ships.get(1) == 4,
+            count_ships.get(2) == 3,
+            count_ships.get(3) == 2,
+            count_ships.get(4) == 1,
+            self.check_neighbours_cells(ships)
+        )
```
Dimosphen1: Do not place checks in a separate variable here, write them in `all` with the use of a list.

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+
+        if all(checks):
+            return True
+        return False
```
Dimosphen1: You can write `return all([...])` where `...` will be your checks. It will return the boolean `True` or `False`. There is no need to return explicitly True or False.

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+            location: tuple,
+            row_and_column: tuple
+    ) -> bool:
+        curr_location = (
+            location[0] + row_and_column[0],
+            location[1] + row_and_column[1]
+        )
+
+        checks = curr_location in self.field, curr_location not in locations
+        return not all(checks)
```
Dimosphen1: Place your checks inside `all`, there is no need for a separate variable here as well.

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+
+    @staticmethod
+    def create_decks_instances(
+            row: int,
+            start: int,
+            end: int,
+            is_row: bool = False
+    ) -> list[Deck]:
+        decks = []
+        for i in range(start, end + 1):
```
Dimosphen1: Do not use variable names like `i` in your code, consider more descriptive naming.

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+            if self.start[1] < self.end[1]:
+                return self.create_decks_instances(
+                    row=self.start[0],
+                    start=self.start[1],
+                    end=self.end[1],
+                    is_row=True
+                )
+
+            if self.start[1] == self.end[1]:
+                return [(Deck(row=self.start[0], column=self.start[1]))]
```
Dimosphen1: There is no need to add `Deck` instance to the tuple here.

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+                    lambda row_and_column: self.check_cell(
+                        locations,
+                        location,
+                        row_and_column
+                    ),
+                    iter(combinations)
+                )
+
+                for _ in not_allowed_cells:
+                    return False
```
Dimosphen1: Rewrite this part with the use of `any`, like `if any(not_allowed_cells): return False`

app/main.py
```diff
@@ -1,34 +1,215 @@
...
+        return False
+
+    def check_neighbours_cells(self, ships: set[Ship]) -> bool:
+        for ship in ships:
+            locations = []
+            for deck in ship.decks:
+                locations.append((deck.row, deck.column))
+
+            for location in locations:
+                combinations = [
```
Dimosphen1: You can declare a variable `combinations` at the beginning of a function and not rewrite it with each iteration.

Dimosphen1 (changes requested):
Several minor changes were requested.

Polyakiv-Andrey (approved):
Good job!

PR #236, opened 2023-07-06
solution + extra tasks

sbilikepy:
i'll do `print_field `and `validate_input `later
skip it for now please


app/main.py
```diff
@@ -1,34 +1,143 @@
...
+                          in range(self.start[1], self.end[1] + 1)]
+            # self.len = len(self.decks)
+            # print(self.len)
+
+        """↓"""
+        if self.direction == "y":
+            self.decks = [Deck(row=coord, column=self.start[1]) for coord
+                          in range(self.start[0], self.end[0] + 1)]
+            # self.len = len(self.decks)
+            # print(self.len)
```
ArsenPidhoretskyi: So hard code....
Just do loop inside loop

app/main.py
```diff
@@ -1,34 +1,143 @@
...
+
+    def get_axis(self):  # custom
```
ArsenPidhoretskyi: This is redundant

app/main.py
```diff
@@ -1,34 +1,143 @@
...
+            self.direction = "single_point"  # □
+            return
+        if self.start[0] == self.end[0]:
+            self.direction = "x"  # □ □ □
+            return
+        self.direction = "y"  # □
+        # ______________________□
+        # ______________________□
+
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,143 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
+        # Find the Deck instance by coords
+
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            self.decks.remove(deck)
```
ArsenPidhoretskyi: No need to remove deck

app/main.py
```diff
@@ -1,34 +1,143 @@
...
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list[tuple]):  # Whole game
+        self.field = {}  # Create a dict `self.field`. # ((), (), ... ()) : self.Ship
+        counter = 1
+        for i in ships:  # debug
```
ArsenPidhoretskyi: i is not good name

app/main.py
```diff
@@ -1,34 +1,143 @@
...
     def fire(self, location: tuple):
+        for coord, ship in self.field.items():
+            point = (coord.row, coord.column)
+            if point == location:
+                ship.fire(*location)
+                if ship.is_drowned is True:
+                    return "Sunk!"
+                else:
+                    return "Hit!"
+        return "Miss!"
```
ArsenPidhoretskyi: Loop is not needed. Just check:
if location in self.field:
.....

ArsenPidhoretskyi (changes requested):

app/main.py
sbilikepy: how can i check it without iteration if my key : value pairs looks like:
`<__main__.Deck object at 0x00000231EA4C2B30>: <__main__.Ship object at 0x00000231EA4C2B00>`
I tried:
`deck = Deck(*location)` but
`self.field[deck]` raised `Keyerror`, even with same hash of deck obj 
`hash((self.row, self.column, self.is_alive))`

app/main.py
```diff
@@ -1,34 +1,233 @@
...
+            return
+
+        self.decks = [Deck(row=coord, column=self.start[1]) for coord
+                      in range(self.start[0], self.end[0] + 1)]
+
+    def get_deck(self, row: int, column: int) -> Union[Deck, None]:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
```
Grayder0152: remove this line

app/main.py
```diff
@@ -1,34 +1,233 @@
...
+            self.decks = [
+                Deck(row=self.start[0], column=coord)
+                for coord in range(self.start[1], self.end[1] + 1)
+            ]
+            return
+
+        self.decks = [Deck(row=coord, column=self.start[1]) for coord
+                      in range(self.start[0], self.end[0] + 1)]
+
+    def get_deck(self, row: int, column: int) -> Union[Deck, None]:
```
```suggestion
    def get_deck(self, row: int, column: int) -> Optional[Deck]:
```

app/main.py
```diff
@@ -1,34 +1,233 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
+
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
```
```suggestion
        if deck is not None:
```

app/main.py
```diff
@@ -1,34 +1,233 @@
...
+            if point == location:
+                ship.fire(*location)
+                if ship.is_drowned is True:
+                    print("Sunk!")
+                    self.print_field()
+                    return "Sunk!"
+                else:
+                    print("Hit!")
+                    self.print_field()
+                    return "Hit!"
```
```suggestion
                    print("Hit!")
                    self.print_field()
                    return "Hit!"
```

Grayder0152 (changes requested):

app/main.py
sbilikepy: i hope you mean "return None" 😄

Grayder0152 (approved):

PR #252, opened 2023-07-28
Solution

laskarj:


app/main.py
```diff
@@ -1,34 +1,125 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        elif self.start[1] == self.end[1]:
+            return [
+                Deck(row=row, column=self.start[1])
+                for row in range(self.start[0], self.end[0] + 1)
+            ]
```
ArsenPidhoretskyi: Just use loop inside loop

app/main.py
```diff
@@ -1,34 +1,125 @@
...
+        raise ValueError("Location entered incorrectly!")
+
+    def get_deck(self, row: int, column: int) -> Deck | str:
```
ArsenPidhoretskyi: It is impossible to return str, but it can return None

app/main.py
```diff
@@ -1,34 +1,125 @@
...
+                )
+            self.limits[ship.size()] -= 1
+            self._range_ship(ship)
+            self.field.update({deck.get_coord(): ship for deck in ship.decks})
+
+    def fire(self, location: Tuple[int, int]) -> str:
+        try:
+            return self.field[location].fire(*location)
+        except KeyError:
+            return "Miss!"
```
```suggestion
        if location in self.field[location]:
            return self.field[location].fire(*location)
        return "Miss!"
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,115 @@
...
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks: List[Deck] = self._create_decks()
+
+    def _create_decks(self) -> list:
```
Grayder0152: list of what ?

app/main.py
```diff
@@ -1,34 +1,115 @@
...
+        return [Deck(row, column)
+                for row in range(self.start[0], self.end[0] + 1)
+                for column in range(self.start[1], self.end[1] + 1)]
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
Grayder0152: add Optional type

app/main.py
```diff
@@ -1,34 +1,115 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def size(self) -> int:
+        return len(self.decks)
+
+    def fire(self, row: int, column: int) -> str:
+        self.get_deck(row, column).is_alive = False
```
Grayder0152: What if `get_deck` return None ?

Grayder0152 (changes requested):

vsmutok (approved):
Good Job 👍

PR #257, opened 2023-08-07
Solution

pnakongit:


app/main.py
```diff
@@ -1,34 +1,96 @@
...
+    @classmethod
+    def create_decks(cls, start: tuple, end: tuple) -> List[Deck]:
+        coordinates_len = end[0] - start[0] + end[1] - start[1] + 1
+        rows = [start[0]] * coordinates_len
+        columns = [start[1]] * coordinates_len
+
+        if start[0] != end[0]:
+            rows = [num for num in range(start[0], end[0] + 1)]
+        elif start[1] != end[1]:
+            columns = [num for num in range(start[1], end[1] + 1)]
```
ArsenPidhoretskyi: This all checks is redundant just use loop inside loop:
for row in range(start[0], start[1]):
    for column in ...

app/main.py
```diff
@@ -1,34 +1,96 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            if all(not deck.is_alive for deck in self.decks):
+                self.is_drowned = True
```
```suggestion
            self.is_drowned = all(not deck.is_alive for deck in self.decks)
```

ArsenPidhoretskyi (changes requested):

app/main.py
pnakongit: Done.

app/main.py
pnakongit: Done

ArsenPidhoretskyi (approved):
Well done!

PR #264, opened 2023-08-24
Solution

tanya-burdenko:


app/main.py
```diff
@@ -1,34 +1,77 @@
...
+                index += 1
+        elif start[1] == end[1] and start[0] != end[0]:
+            index = 0
+            for _ in range(start[0], end[0] + 1):
+                self.decks.append(Deck(start[0] + index,
+                                       start[1],
+                                       is_alive=True))
+                index += 1
+        else:
+            self.decks.append(Deck(start[0], start[1], is_alive=True))
```
ArsenPidhoretskyi: Use loop inside loop instead of if statements

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+        self.decks = []
+        if start[0] == end[0] and start[1] != end[1]:
+            index = 0
+            for _ in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0],
+                                       start[1] + index,
+                                       is_alive=True))
+                index += 1
+        elif start[1] == end[1] and start[0] != end[0]:
+            index = 0
```
ArsenPidhoretskyi: Index variable is not needed. Result will be the same if you instead _ will use index

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        if start[0] == end[0] and start[1] != end[1]:
+            index = 0
+            for _ in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0],
+                                       start[1] + index,
+                                       is_alive=True))
```
ArsenPidhoretskyi: No need to provide default argument

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+        if deck:
+            deck.is_alive = False
+        check = []
+        for deck in self.decks:
+            if not deck.is_alive:
+                check.append(True)
+            else:
+                check.append(False)
+        if all(check):
+            self.is_drowned = True
```
ArsenPidhoretskyi: Write this inline
self.is_drowned = all(....)

app/main.py
```diff
@@ -1,34 +1,77 @@
...
-        # in the ship or not.
-        pass
+        self.ships = []
+        for ship in ships:
+            self.ships.append(Ship(ship[0], ship[1], is_drowned=False))
+
+        self.field = {}
+        for ship in self.ships:
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
```
ArsenPidhoretskyi: Use dict-comprehension

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,56 @@
...
+        self.decks = [
+            Deck(row, column)
+            for row in range(start[0], end[0] + 1)
+            for column in range(start[1], end[1] + 1)
+        ]
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,56 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+        check = [True if not deck.is_alive else False for deck in self.decks]
+        self.is_drowned = True if all(check) else False
```
ArsenPidhoretskyi: First of all you can do:
self.is_drowned = all(check)

app/main.py
```diff
@@ -1,34 +1,56 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+        check = [True if not deck.is_alive else False for deck in self.decks]
```
ArsenPidhoretskyi: The same here no need to write if true else false

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,59 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+        check = [not deck.is_alive for deck in self.decks]
+        self.is_drowned = all(check)
```
```suggestion
        self.is_drowned = all(not deck.is_alive for deck in self.decks)
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #265, opened 2023-08-24
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        return len(self.decks)
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self,
+                 row: int,
+                 column: int
+                 ) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+             column: int
+             ) -> None:
+        target_deck = self.get_deck(row, column)
+        target_deck.is_alive = False
+        is_drowned = True
+        for deck in self.decks:
+            if deck.is_alive:
+                is_drowned = False
+                break
+        self.is_drowned = is_drowned
```
ArsenPidhoretskyi: Use all or any

ArsenPidhoretskyi (changes requested):

PR #270, opened 2023-08-26
Implemented py-battleship

esemesenko:


app/main.py
```diff
@@ -1,34 +1,160 @@
...
+            for cell in cells:
+                row, column = cell
+                if (row, column) in empty_cells_set:
+                    raise _FieldError("Ships shouldn't "
+                                      "be located in the neighboring cells")
+
+    @staticmethod
+    def _add_new_cell(
+            cells: tuple,
+            checked_cells: list,
```
Bohdan-Salamakha: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #275, opened 2023-08-29
Solution

AlexPelevin99:


app/main.py
```diff
@@ -1,34 +1,187 @@
...
+        self.length = int(math.sqrt(diff_row ** 2 + diff_column ** 2) + 1)
+
+        self.decks: List[Deck] = self.__create_decks()
+
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+        return None
```
Dimosphen1: Remove this `return` from the code, `None` is a default return value.

app/main.py
```diff
@@ -1,34 +1,187 @@
...
+        deck = self.get_deck(row, column)
+        if deck.is_alive:
+            deck.is_alive = False
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        if not self.is_drowned and sum(
+                [deck.is_alive for deck in self.decks]
```
Dimosphen1: Do not use list comprehension inside of sum, remove `[` and `]`.

Dimosphen1: You could use `any()`  here as well.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #279, opened 2023-08-31
Solution

AndriiIshchenko:


app/main.py
```diff
@@ -1,34 +1,86 @@
...
-        # Find the corresponding deck in the list
-        pass
+        if start == end:
+            self.decks = [Deck(*start), ]
+        if start[0] == end[0]:
+            self.decks = [Deck(start[0], i)
+                          for i in range(start[1], end[1] + 1)]
+        if start[1] == end[1]:
+            self.decks = [Deck(i, start[1])
+                          for i in range(start[0], end[0] + 1)]
```
ArsenPidhoretskyi: Why do you need to use if statement instead just use for-loop

app/main.py
```diff
@@ -1,34 +1,86 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,86 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, target: Deck) -> str:
+        target.is_alive = False
+        if sum([deck.is_alive for deck in self.decks]) == 0:
+            self.is_drowned = True
```
```suggestion
        self.is_drowned = not any([deck.is_alive for deck in self.decks])
```

app/main.py
```diff
@@ -1,34 +1,86 @@
...
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {
+            tuple([(deck.row, deck.column) for deck in Ship(*coord).decks]):
+            Ship(*coord) for coord in ships
+        }
```
ArsenPidhoretskyi: You create each Ship twice. Replace comprehensions by forloop

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiIshchenko: I think it can't return None i call  **get_deck** only after cheking
    def fire(self, location: tuple) -> str:
        for coords, ship in self.field.items():
            if location in coords:
                return ship.fire(ship.get_deck(*location))

get_deck just shooses the right deck from current ship

the same in print_field i use get_deck after cheking

app/main.py
AndriiIshchenko: I'm just overthinking this :)

app/main.py
```diff
@@ -1,34 +1,83 @@
...
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for coords in ships:
+            ship = Ship(*coords)
+            self.field.update({tuple(
+                [(deck.row, deck.column) for deck in ship.decks]
+            ): ship})
+
+    def fire(self, location: tuple) -> str:
```
```suggestion
    def fire(self, location: tuple[int]) -> str:
```

app/main.py
vasylhnatiuk: Yes, but this function can return None. We should not think about where this function is used now; we must predict how this function can be used in different cases.

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,86 @@
...
-        # which is located in it
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
```
AndriiIshchenko, line 33: here I must use List[tuple] or not? I think not, we dont know about incomming data and we just do not chek it in this task

vasylhnatiuk (approved):

PR #285, opened 2023-09-01
Solution

denys-source:


app/main.py
```diff
@@ -1,34 +1,96 @@
...
+                return deck
+        return None
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        deck: Deck | None = self.get_deck(row, column)
+        if deck is not None:
```
Kroll410: simple if deck, because Deck is a dataclass

app/main.py
```diff
@@ -1,34 +1,96 @@
...
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(
+        self, ships: list[tuple[tuple[int, int], tuple[int, int]]]
```
Kroll410, line 44: I'd suggest using TypeVar here, because this typing might be correct, but I don't understand what is there

app/main.py
```diff
@@ -1,34 +1,96 @@
...
+
+        for start, end in ships:
+            ship: Ship = Ship(start, end)
+            for coord in Battleship.generate_coordinates(start, end):
+                self.field[coord] = ship
+
+    def fire(self, location: tuple[int, int]) -> str:
+        if (ship := self.field.get(location, None)) is not None:
+            ship.fire(*location)
+            if ship.is_drowned:
```
Kroll410: Use ternary

app/main.py
```diff
@@ -1,34 +1,96 @@
...
+        return "Miss!"
+
+    def print_field(self) -> None:
+        result: list[str] = []
+        row: list[str] = []
+        for row_coord in range(10):
+            for col_coord in range(10):
+                if (
+                    ship := self.field.get((col_coord, row_coord), None)
+                ) is not None:
```
Kroll410: None is being send by default from get, no need to specify

Kroll410: no need to pass here `is not None`

app/main.py
```diff
@@ -1,34 +1,96 @@
...
+                    if ship.is_drowned:
+                        row.append("x")
+                    else:
+                        deck: Deck | None = ship.get_deck(col_coord, row_coord)
+                        if deck and deck.is_alive:
+                            row.append("\u25A1")
```
Kroll410: Use Ternary

Kroll410 (changes requested):
Gj, but there's room to improve

app/main.py
denys-source: @Kroll410 could you, please, give me some hints on how to do that

WonnaKoo (approved):

IvanRamyk (approved):

PR #299, opened 2023-09-26
Solution: Battleship.

lobodInI:


app/main.py
```diff
@@ -1,34 +1,82 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row=row, column=column)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+        if deck is not None:
+            deck.is_alive = False
+            self.decks.remove(deck)
```
viktoria-rybenchuk: Use 'if deck' instead of 'if deck is not None' for a more concise and Pythonic approach

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #301, opened 2023-09-27
Implemented Battleship

mtiunov:


app/main.py
```diff
@@ -1,34 +1,78 @@
...
+    for ship in battle_ship.ships:
+        for deck in ship.decks:
+            field[deck[0]][deck[1]] = ship
+
+    for row in range(10):
+        for col in range(10):
+            if field[row][col] is None:
+                field[row][col] = Deck(row, col)
+
+    battle_ship.print_field(field)
```
YukihiroSM: Remove this testing code block.

tests/test_validate.py
YukihiroSM, line 1: What is this test for?

YukihiroSM (changes requested):

tests/test_validate.py
mtiunov: This test for extra tasks: validate input

YukihiroSM: Then it should be a part of class. If you read the task: 
Write protected method _validate_field that should check the following conditions after creating a field in the __init__ method
It says, that you need to write protected method - it's only for the class. Also, it should be executed in __init__ to validate input

YukihiroSM (changes requested):
Looks into the previous comments

mtiunov:
> Looks into the previous comments


I have to delete this file > tests/test_validate.py ?

mtiunov:
Sorry, I just saw your comment.

app/main.py
```diff
@@ -1,34 +1,47 @@
...
-        # which is located in it
-        pass
+    def fire(self, location: tuple) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            hit = ship.fire(*location)
+            if hit:
+                if ship.is_drowned:
+                    return "Sunk!"
+                else:
```
Bohdan-Salamakha: remove redundant 'else'

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #306, opened 2023-10-02
solution

ostboiko:


app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        self._create_decks()
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def _create_decks(self) -> None:
```
ArsenPidhoretskyi: This method is redundant just do the same into init

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        ]
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
ArsenPidhoretskyi, line 27: What about fire method? Why you remove it?

ArsenPidhoretskyi (changes requested):

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

NikOneZ1 (changes requested):
Сomments left by previous reviewers were not fixed

ostboiko:
Hello everyone, I moved the fire function to another class, but I'm not sure what's right, can you suggest a further solution to the problem

app/main.py
ArsenPidhoretskyi: not fixed

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+            for row in range(self.start[0], self.end[0] + 1)
+            for column in range(self.start[1], self.end[1] + 1)
+        ]
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, location: tuple[int]) -> str:
+        if location in self.field:
```
ArsenPidhoretskyi: Really self.field exist into Ship???

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Why you resolve it when it is still bug

app/main.py
```diff
@@ -1,34 +1,89 @@
...
+    def _create_field(self, ships: list[int]) -> None:
+        for ship_start, ship_end in ships:
+            ship = Ship(ship_start, ship_end)
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
+
+    def fire(self, location: tuple[int]) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            deck = ship.get_deck(*location)
```
ArsenPidhoretskyi: Use 
```suggestion
            ship.fire(*location)
```

ArsenPidhoretskyi: Use fire from Ship method

ArsenPidhoretskyi (changes requested):

app/main.py
ostboiko: sorry didn't see this message

EdAlekseiev (approved):
Good job

PR #309, opened 2023-10-06
'Solution'

jinjo0222988:


app/main.py
```diff
@@ -1,34 +1,106 @@
...
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self,
+                 start: Tuple[int, int],
+                 end: Tuple[int, int],
+                 is_drowned: bool = False) -> None:
+        self.is_drowned = is_drowned
+        self.decks = Ship.create_decks(start, end)
+
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: It can also return None

app/main.py
```diff
@@ -1,34 +1,106 @@
...
+        if self.is_drowned:
+            return "Sunk!"
+        else:
+            return "Hit!"
+
+    def is_it_drowned(self) -> None:
+        self.is_drowned = True
+        for deck in self.decks:
+            if deck.is_alive:
+                self.is_drowned = False
```
ArsenPidhoretskyi: Just use any/all

app/main.py
```diff
@@ -1,34 +1,106 @@
...
+        border = 0
+        if ship_begin[1] != ship_end[1]:
+            border = 1
+
+        ship_decks = [Deck(ship_begin[0], ship_begin[1])]
+        blank_deck = [ship_begin[0], ship_begin[1]]
+
+        while blank_deck[border] != ship_end[border]:
+            blank_deck[border] += 1
+            ship_decks.append(Deck(blank_deck[0], blank_deck[1]))
```
ArsenPidhoretskyi: Instead of if+ while just use forloop into forloop

app/main.py
```diff
@@ -1,34 +1,106 @@
...
+            temp_ship = Ship(*ship)
+            for deck in temp_ship.decks:
+                self.field[deck.coordinate] = temp_ship
+
+    def fire(self, location: Tuple[int, int]) -> str:
+        ship = self.field.get(location)
+        if ship is not None:
+            return ship.fire(*location)
+        else:
+            return "Miss!"
```
```suggestion
        return "Miss!"
```

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: not fixed

app/main.py
```diff
@@ -1,34 +1,98 @@
...
+
+    @staticmethod
+    def create_decks(ship_begin: Tuple[int, int],
+                     ship_end: Tuple[int, int]) -> List:
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+        if ship_begin == ship_end:
+            return [Deck(*ship_begin)]
```
ArsenPidhoretskyi: This check is redundant

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #310, opened 2023-10-09
Solution by Dyriv

IvDyriv:


app/main.py
```diff
@@ -1,34 +1,69 @@
...
-        # Find the corresponding deck in the list
-        pass
+    def make_decks_for_ship(self) -> list:
+        decks = []
+        ship_length = self.end[0] - self.start[0] + self.end[1] - self.start[1]
+        for i in range(0, ship_length + 1):
+            if self.start[0] == self.end[0]:
+                decks.append(Deck(self.start[0], self.start[1] + i))
+            else:
+                decks.append(Deck(self.start[0] + i, self.start[1]))
```
ArsenPidhoretskyi: Instead of some calculation and if statements just use for loop

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        return decks
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```
ArsenPidhoretskyi: It also can return None

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        for deck in self.decks:
+            if row == deck.row and column == deck.column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
+        self.is_drowned = True
+        for deck in self.decks:
+            if deck.is_alive:
+                self.is_drowned = False
```
ArsenPidhoretskyi: Use any/all

ArsenPidhoretskyi (changes requested):

app/main.py
IvDyriv: Thank you

app/main.py
IvDyriv: Thank you

app/main.py
IvDyriv: Thank you

app/main.py
```diff
@@ -1,34 +1,70 @@
...
+    def make_decks_for_ship(self) -> list:
+        decks = []
+        if self.start[0] == self.end[0]:
+            # Ship is oriented horizontally
+            for column in range(self.start[1], self.end[1] + 1):
+                decks.append(Deck(self.start[0], column))
+        else:
+            # Ship is oriented vertically
+            for row in range(self.start[0], self.end[0] + 1):
+                decks.append(Deck(row, self.start[1]))
```
ArsenPidhoretskyi: Use loop into loop instead of if statement

app/main.py
ArsenPidhoretskyi: Fix typing in this case

app/main.py
```diff
@@ -1,34 +1,70 @@
...
+    def __init__(self,
+                 ships: list
+                 ) -> None:
+        self.ships = ships
+        self.field = self.making_ships()
+
+    def making_ships(self) -> dict:
+        field = {}
+        for ship in self.ships:
+            ship1 = Ship(ship[0], ship[1])
```
```suggestion
            ship_obj = Ship(ship[0], ship[1])
```

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #315, opened 2023-10-13
Solution

olgierrd:
my _validate_field method is not complete, as I could not find a good way of checking whether the ships are neighbors, but it was additional anyway :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,34 +1,90 @@
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(self, row: int, column: int, is_alive: bool = True) -> None:
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        length = abs(start[0] - end[0]) + abs(start[1] - end[1]) + 1
+        self.decks = []
+        for i in range(length):
+            if start[0] == end[0]:
+                deck = Deck(start[0], start[1] + i)
+            else:
+                deck = Deck(start[0] + i, start[1])
+            self.decks.append(deck)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,34 +1,90 @@
...
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        length = abs(start[0] - end[0]) + abs(start[1] - end[1]) + 1
+        self.decks = []
+        for i in range(length):
```
pavlejviki: don't use i as variable name

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+                        if ship.get_deck(row, column).is_alive:
+                            print(u"\u25A1", end="  ")
+                        else:
+                            print("*", end="  ")
+                else:
+                    print("~", end="  ")
+            print()
+
+    def _validate_field(self) -> None:
+        n_ships = len(self.ships)
```
pavlejviki: n_ships is not a clear variable name

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+                    if ship.is_drowned:
+                        print("x", end="  ")
+                    else:
+                        if ship.get_deck(row, column).is_alive:
+                            print(u"\u25A1", end="  ")
+                        else:
+                            print("*", end="  ")
+                else:
+                    print("~", end="  ")
+            print()
```
pavlejviki, line 72: redundant print

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+                self.field[(deck.row, deck.column)] = ship
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            ship.fire(location[0], location[1])
+            if ship.is_drowned:
+                return "Sunk!"
+            return "Hit!"
+        else:
```
pavlejviki: redundant else

pavlejviki (changes requested):
It's better to completely remove an optional method if it's not functioning properly, rather than keep it in a dysfunctional state.

app/main.py
olgierrd: this print jumps to a new row, so the result is 10x10 grid instead of 1x100 row

app/main.py
viktoria-rybenchuk: not fixed

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+            self.decks.append(deck)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
```
viktoria-rybenchuk: If no deck is found, the function automatically returns None.

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #321, opened 2023-10-24
Solution

D1st3f:


Naz-iv (approved):

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+            self.decks = [Deck(i, self.start[1])
+                          for i in range(self.start[0], self.end[0] + 1)]
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
wQuelS: can this method get the wrong location? If yes - your return type annotation is not correct

app/main.py
```diff
@@ -1,34 +1,61 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS: If `get_deck` can return None in cases when not valid location is passed, it is very unsafe to use `is_alive` attr on the object that potentially can be `None`

app/main.py
```diff
@@ -1,34 +1,61 @@
...
+
+        drowned = []
+
+        for deck in self.decks:
+            drowned.append(deck.is_alive)
+        if not any(drowned):
+            self.is_drowned = True
```
wQuelS: this can be simplified by using generator expression along with `any`

app/main.py
```diff
@@ -1,34 +1,61 @@
...
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.ships = ships
+        self.field = {}
+        for start, end in self.ships:
+            ship = Ship(start, end)
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
```
wQuelS, line 55: I think, it is better to move such "creational" logic to a separate method, that can be called in `init`. Just like you've done in the `Ships` class

wQuelS (changes requested):

app/main.py
```diff
@@ -27,26 +28,27 @@ def fill_decs(self) -> None:
             self.decks = [Deck(i, self.start[1])
                           for i in range(self.start[0], self.end[0] + 1)]
 
-    def get_deck(self, row: int, column: int) -> Deck:
+    def get_deck(self, row: int, column: int) -> Optional[Deck]:
         for deck in self.decks:
             if deck.row == row and deck.column == column:
                 return deck
+        return None
```
wQuelS, line 35: just to mention, that this explicit `return None` is not needed, because if no `Deck` is found - it will returned by default, because no `return` statement will be reached

wQuelS (approved):
Gj, left one comment as a suggestion. But everything looks good!

PR #322, opened 2023-10-25
Solution

Naz-iv:
Implemented _validate method but it's not used anywhere yet, as I'm not sure when validation should be performed. Also validate fields method is missing check for neighboring ships.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -9,9 +9,6 @@ class Deck:
         self.column = column
         self.is_alive = is_alive
 
-    def __repr__(self) -> str:
-        return f"Deck: {self.row} : {self.column}"
-
     def __str__(self) -> str:
         return f"Deck: {self.row} : {self.column}"
 
@@ -28,13 +25,11 @@ class Ship:
             if deck.row == row and deck.column == column:
                 return deck
 
-    def fire(self, row: int, column: int) -> bool:
+    def fire(self, row: int, column: int) -> None:
         deck = self.get_deck(row, column)
         if deck and deck.is_alive:
             deck.is_alive = False
-        decks_available = [deck for deck in self.decks if deck.is_alive]
-        if not decks_available:
-            self.is_drowned = True
+        self.is_drowned = not any(deck.is_alive for deck in self.decks)
 
     @staticmethod
     def _decks_generator(start: tuple, end: tuple) -> list:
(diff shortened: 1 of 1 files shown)


prochigor (approved):

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+    def __init__(self, row: int, column: int, is_alive: bool = True) -> None:
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
+
+    def __repr__(self) -> str:
+        return f"Deck: {self.row} : {self.column}"
+
+    def __str__(self) -> str:
+        return f"Deck: {self.row} : {self.column}"
```
wQuelS: You've added these methods despite the fact that the task was not even asking you to do so. - That's good
But here I can see some problems. Both methods are identical. As for `__str__` - such implementation is pretty much okay, 
My concern is mainly addressed to `__repr__`. The goal of `__repr__` is to be unambiguous, meaning that it  is supposed to return an "official" string representation of an object. The idea is that this should be, as much as possible, a string that could be used to recreate the object if passed to `eval()`.

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> bool:
+        deck = self.get_deck(row, column)
+        if deck and deck.is_alive:
+            deck.is_alive = False
+        decks_available = [deck for deck in self.decks if deck.is_alive]
+        if not decks_available:
+            self.is_drowned = True
```
wQuelS: I bet this can be done with `any` and generator expression, so you will not need to create a list with decks that are alive

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+                four += 1
+
+        if single != 1:
+            return "You should have 4 single-deck ships"
+        if double != 2:
+            return "You should have 3 double-deck ships"
+        if three != 3:
+            return "You should have 2 three-deck ships"
+        if four != 4:
+            return "You should have 1 four-deck ships"
```
wQuelS: based on the messages, It seems like a logical mistake here, doesn't it? 

Also, by fixing code in that is above this part -  you would be able to use that dictionary to simplify your code by removing numerous if statements  ( details are in dedicated comment )

app/main.py
```diff
@@ -1,34 +1,113 @@
...
-        # in the ship or not.
-        pass
+
+    def __init__(self, ships: List[Ship]) -> None:
+        self.field = self.get_field(ships)
+        self.draw_playfield()
+
+    def _validate_fields(self) -> str:
+        ships = self.field.values()
+        if set(ships) != 10:
```
wQuelS: It will always return `False` :)

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+        single, double, three, four = 0, 0, 0, 0
+        for ship in set(ships):
+            if len(ship.decks) == 1:
+                single += 1
+            if len(ship.decks) == 2:
+                double += 1
+            if len(ship.decks) == 3:
+                three += 1
+            if len(ship.decks) == 4:
+                four += 1
```
wQuelS: 1) On each iteration every `if` condition is gonna be checked. Even if length of deck is 1, and next checks are not needed.
2) A fix with `elif` is not that good also. You can optimize and improve the logic for checking the quantity of ships. Instead of manually counting the number of ships for each length, you can use a dictionary to store the counts dynamically. A dictionary can have both key and value as ints that would refer to the number of decks and quantity of such ships respectively. This makes the code more flexible and avoids repetitive conditions.

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+        if three != 3:
+            return "You should have 2 three-deck ships"
+        if four != 4:
+            return "You should have 1 four-deck ships"
+
+    def fire(self, location: tuple) -> str:
+        ship = self.field.get(location)
+
+        if ship:
+            self.field.pop(location)
```
wQuelS: Your ships are no joke, they are evaporating enemies on hit...

By making ships disappear instead of setting them as killed you have established one more problem. For example, if a target ship has three decks, on this line it completely loses a deck, and now it is impossible to say which ship those two remaining docks belong to. Because at this moment in a `field` dictionary there is an additional two-deck ship

app/main.py
```diff
@@ -1,34 +1,113 @@
...
+
+        return field
+
+    def draw_playfield(self) -> None:
+        self.playfield = [["~" for _ in range(10)] for _ in range(10)]
+        for deck in self.field.keys():
+            row, column = deck
+            self.playfield[row][column] = "□"
+
+        print(self.playfield)
```
wQuelS: This method should return play field in its current state. Meaning that if some ship is damaged or even done - playfield should show that. For now, your method is useful only to show the initial state, because damaged `*` or killed `X` ships, that you have from fire method are overwritten here. Also, your play field looks not that good, and it is hard to read

image
This screenshot shows how your field is rendered after some ship is killed. Notice, that instead of showing killed ship, your battlefield renders `~~~~` in its place.

image
This screenshot is an example, of how the field should look like after a ship is sunk.

wQuelS (changes requested):

app/main.py
Naz-iv: Only added them for troubleshooting and forgot to remove before committing

app/main.py
Naz-iv: good call, thanks, wrote it with any

app/main.py
Naz-iv: followed your advice and got way better solution, thanks

app/main.py
Naz-iv: I added this check since your validation method should check if number of shits is exactly 10 and since multiple key: value pairs may have reference to same ship I had to make set of it

app/main.py
Naz-iv: followed approach with dictionary, thanks for advice

app/main.py
Naz-iv: hah, no clue why I need it in first place))) removed now

app/main.py
Naz-iv: created additional private method that initializes  play field and another method that prints out filed in user friendly format

LiudmylaKulinchenko (approved):
Well done ☀

PR #323, opened 2023-10-25
Solution

prochigor:


vshvanska (approved):

app/main.py
```diff
@@ -1,34 +1,49 @@
...
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        self.decks = []
+        for row in range(start[0], end[0] + 1):
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(row, column))
```
wQuelS, line 19: I'd suggest creating a separate method which will create decks and add them to a list.

app/main.py
```diff
@@ -1,34 +1,49 @@
...
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.field = {}
+        for ship in ships:
+            created_ship = Ship(ship[0], ship[1])
+            for deck in created_ship.decks:
+                self.field[(deck.row, deck.column)] = created_ship
```
wQuelS, line 40: and again, such "creational" logic should be in separate method which can be called from `init`

app/main.py
```diff
@@ -1,34 +1,49 @@
...
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def fire(self, location: tuple) -> str:
+        if location not in self.field:
+            return "Miss!"
+        ship = self.field[location]
+        ship.fire(location[0], location[1])
```
wQuelS, line 46: do you really need to pass each element from the tuple separately? Why not just pass whole tuple to `fire` method ?

wQuelS (changes requested):

PR #324, opened 2023-10-25
Solution

vshvanska:


YuliiaSaliuta (approved):

app/main.py
```diff
@@ -1,34 +1,50 @@
...
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        (start_r, start_c), (end_r, end_c) = start, end
+        for row in range(start_r, end_r + 1):
+            for column in range(start_c, end_c + 1):
+                deck = Deck(row, column)
+                self.decks.append(deck)
```
wQuelS, line 21: it is better to move this logic to separate method, which you can call from `init`

app/main.py
```diff
@@ -1,34 +1,50 @@
...
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for ship in ships:
+            current_ship = Ship(*ship)
+            for deck in current_ship.decks:
+                self.field[deck.row, deck.column] = current_ship
```
wQuelS, line 41: again, such "creational" logic better suits for some method

wQuelS (changes requested):

PR #325, opened 2023-10-25
Solution

YuliiaSaliuta:


RVChornyy (approved):

app/main.py
```diff
@@ -1,34 +1,61 @@
...
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, location: tuple) -> None:
+        row, column = location
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
```
wQuelS: Can your `get_deck` return None if there is no deck at that location?
If yes, do you think it is safe to use `is_alive` on the object, that potentially can be None?

app/main.py
```diff
@@ -1,34 +1,61 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for ship in ships:
+            my_ship = Ship(ship[0], ship[1])
+            for deck in my_ship.decks:
+                self.field[(deck.row, deck.column)] = my_ship
```
wQuelS, line 55: such logic better suited for some separate method, that can be called from this `init`

wQuelS (changes requested):

wQuelS (approved):
Well done :fire:

PR #326, opened 2023-10-25
Develop

RVChornyy:


app/main.py
```diff
@@ -1,34 +1,69 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.column == column and deck.row == row:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS, line 48: can your `get_deck` return None if there is no ship at that location? If yes, do you think it is safe to use `is_alive` attr on the object that potentially can be None?

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        self.is_drowned = not any(list(deck.is_alive for deck in self.decks))
```
wQuelS, line 49: do you really need that `list()` here?

app/main.py
```diff
@@ -1,34 +1,69 @@
...
-        pass
+    def __init__(self, ships: list) -> None:
+        ships = validate_field(ships)
+        if not ships:
+            raise ValueError("Enter correct ships list")
+        self.field = {}
+        ship_list = [Ship(ship[0], ship[1]) for ship in ships]
+        for ship in ship_list:
+            for deck in ship.decks:
+                self.field[deck.row, deck.column] = ship
```
wQuelS, line 61: I think such "creational" logic should be moved to separate method that can be called from `init`. Just like u've done in `Ships` class:)

app/validation.py
```diff
@@ -0,0 +1,46 @@
...
+                        or four_deck > 1):
+                    raise QuantityError
+        except LocationError:
+            print("Ships must be located in rows and columns")
+        except DeckError:
+            print("Ship must contain maximum 4 decks")
+        except QuantityError:
+            print("Too much of this type of boats")
+        else:
+            return ships
```
wQuelS, line 46: This function could be simplified by using a dictionary to count the different types of ships, reducing the number of if statements.
It also lacks checks for the adjacency of ships (e.g., ensuring no ships are placed next to each other diagonally).

Since this is additional - task fixes to this function is up to you

wQuelS (changes requested):

PR #327, opened 2023-10-26
BattleShip solution

Lyutillis:


Familenko (approved):

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+    def __init__(self,
+                 start: tuple,
+                 end: tuple,
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = [Deck(x, y)
+                      for x in range(start[0], end[0] + 1)
+                      for y in range(start[1], end[1] + 1)]
```
wQuelS, line 22: I guess it is better to move such logic to a separate method, which can be called from this `init`

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+        self.length = len(self.decks)
```
wQuelS, line 23: why do you even need it?

app/main.py
```diff
@@ -1,34 +1,90 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
wQuelS, line 28: Is it possible that not a valid location is passed to this method? If yes, than your return type annotation is not correct

app/main.py
```diff
@@ -1,34 +1,90 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        hit_deck = self.get_deck(row, column)
+        hit_deck.is_alive = False
+        if all([not deck.is_alive for deck in self.decks]):
```
wQuelS, line 33: do you really need that `list comprehension` here? Why not just generator expression?

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        hit_deck = self.get_deck(row, column)
+        hit_deck.is_alive = False
```
wQuelS, line 32: If your `get_deck` can return None in cases when not a valid location is passed - it is very unsafe to use `is_alive` attr on the object which can potentially be `None`

app/main.py
```diff
@@ -1,34 +1,90 @@
...
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self._validate_field(ships)
+        self.ships = [Ship(*ship) for ship in ships]
+        self.field = [["~" for _ in range(10)] for _ in range(10)]
```
wQuelS, line 41: again, I'd suggest to separate initialization and creation at least by implementing separate methods that can be called in `init`

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+
+    def _validate_field(self, ships: list) -> None:
+        if len(ships) != 10:
```
wQuelS, line 44: you've added `length` attr to a `Ships` class but you are not using it:)

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+                ships_count[ship[1][1] - ship[0][1] + 1] += 1
+        if not all((ships_count[1] == 4,
+                    ships_count[2] == 3,
+                    ships_count[3] == 2,
+                    ships_count[4] == 1)):
+            raise InvalidInput(("There have to be:\n"
+                                "4 1-deck ships\n"
+                                "3 2-deck ships\n"
+                                "2 3-deck ships\n"
+                                "1 4-deck ship\n"))
```
wQuelS, line 65: What if the number of decks is correct for each ship type, but they are placed at invalid locations?

Just saying, because this method is optional, so fixes are up to you

app/main.py
```diff
@@ -1,34 +1,90 @@
...
+
+    def fire(self, location: tuple) -> str:
+        for ship in self.ships:
+            if all((location[0] in range(ship.start[0], ship.end[0] + 1),
+                    location[1] in range(ship.start[1], ship.end[1] + 1))):
+                ship.fire(*location)
+                if ship.is_drowned:
+                    return "Sunk!"
+                return "Hit!"
+        return "Miss!"
```
wQuelS, line 75: isn't it better to return `"Miss!"` at the very beginning if the location is wrong and if `"Miss!"` is not returned - enter a loop?

wQuelS (changes requested):

PR #329, opened 2023-10-28
sol

Familenko:


petrykivd (approved):

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+
+        self.is_drowned = is_drowned
+        self.decks = []
+
+        x_start, x_end = start[0], end[0]
+        y_start, y_end = start[1], end[1]
+
+        if x_start != x_end:
+            for x_pos in range(x_start, x_end + 1):
+                self.decks.append(Deck(x_pos, start[1]))
```
wQuelS, line 25: I think, such logic should be moved to a separate method, that can be called from this `init`

app/main.py
```diff
@@ -1,34 +1,77 @@
...
-        # And update the `is_drowned` value if it's needed
-        pass
+        else:
+            self.decks.append(Deck(start[0], end[1]))
+
+    def get_deck(self, row: int, column: int) -> Deck:
+
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
wQuelS, line 38: what if there is no `Deck` at the location? If such situation is possible, your return type annotation is wrong

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+
+        self.get_deck(row, column).is_alive = 0
```
wQuelS, line 42: if `get_deck` can return None for cases when there is no ship at given location, it is very unsafe to use `is_alive` attr on a potentially `None` object

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+        self.is_drowned = all([not deck.is_alive for deck in self.decks])
```
wQuelS, line 43: do you really need that `list comprehension here`? Why not just generator expression?

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+            if ship[0][0] != ship[1][0]:
+                for x_pos in range(ship[0][0], ship[1][0] + 1):
+                    self.field[(ship[0][1], x_pos)] = boat
+
+            elif ship[0][1] != ship[1][1]:
+                for y_pos in range(ship[0][1], ship[1][1] + 1):
+                    self.field[(ship[0][0], y_pos)] = boat
+
+            else:
+                self.field[ship[0]] = boat
```
wQuelS, line 63: again, it is better to move such "creational logic" to separate methods

app/main.py
```diff
@@ -1,34 +1,77 @@
...
+            return "Miss!"
+
+        elif location in self.field:
+            self.field[location].fire(location[0], location[1])
+
+            if self.field[location].is_drowned:
+                return "Sunk!"
+
+            else:
+                return "Hit!"
```
wQuelS, line 77: do you really need that `else` here?

wQuelS (changes requested):

PR #330, opened 2023-10-28
Solution

petrykivd:


Kateryna-Bordonos17 (approved):
Well done! 👍

Lyutillis (approved):

app/main.py
```diff
@@ -1,34 +1,56 @@
...
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.ships = ships
+        self.field = {}
+        for start, end in self.ships:
+            ship = Ship(start, end)
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
```
wQuelS, line 52: it is better to move such logic to a separate method

app/main.py
```diff
@@ -1,34 +1,56 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS: Is it possible that the wrong location was given to a `get_deck` method? If yes and it can return `None`, than it is very unsafe to use `is_alive` attr on an object that potentially can be `None`

app/main.py
```diff
@@ -1,34 +1,56 @@
+import dataclasses
+
+
+@dataclasses.dataclass
```
wQuelS: you can import it in a more readable way, so you will be able to use just `@dataclass` instead of `@dataclasses.dataclass`

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #331, opened 2023-10-28
Solution

Kateryna-Bordonos17:


RabenkoYevhenii (approved):
good

app/main.py
```diff
@@ -1,34 +1,64 @@
...
+            self.decks.append(Deck(start[0], end[1]))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
wQuelS: Is it possible that the wrong location is given? If yes - your return type annotation is not correct

app/main.py
```diff
@@ -1,34 +1,64 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS: If `get_deck` can return None in cases when not valid location is given, it is very unsafe to use `is_alive` attr on the object, that potentially is `None`.

app/main.py
```diff
@@ -1,34 +1,64 @@
...
+        self.is_drowned = all([not deck.is_alive for deck in self.decks])
```
wQuelS: do you really need that `list comprehension` here? Why not just generator expression?

app/main.py
```diff
@@ -1,34 +1,64 @@
...
+        for ship in ships:
+            cell = Ship(ship[0], ship[1])
+            if ship[0][0] != ship[1][0]:
+                for x_pos in range(ship[0][0], ship[1][0] + 1):
+                    self.field[(ship[0][1], x_pos)] = cell
+            if ship[0][1] != ship[1][1]:
+                for y_pos in range(ship[0][1], ship[1][1] + 1):
+                    self.field[(ship[0][0], y_pos)] = cell
+            else:
+                self.field[ship[0]] = cell
```
wQuelS, line 58: I think it is better to move such "creational" logic to a separate method.

app/main.py
```diff
@@ -1,34 +1,64 @@
...
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(self,
+                 start: tuple[int],
+                 end: tuple[int],
+                 is_drowned: bool = False) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
```
wQuelS: where are your ships supposed to be created and how?

app/main.py
```diff
@@ -1,34 +1,64 @@
...
+                self.field[ship[0]] = cell
+
+    def fire(self, location: tuple[int]) -> str:
+        if location not in self.field:
+            return "Miss!"
+        else:
+            self.field[location].fire(location[0], location[1])
+            if self.field[location].is_drowned:
+                return "Sunk!"
+            return "Hit!"
```
wQuelS: do you really need this `else` statement here?

wQuelS (changes requested):

app/main.py
Kateryna-Bordonos17: Now I've added a separate method to create ships that generates and returns a list of decks

LiudmylaKulinchenko (approved):
Good job 👏

PR #332, opened 2023-10-29
Solution

RabenkoYevhenii:


Daniil-Pankieiev (approved):

app/main.py
```diff
@@ -1,34 +1,59 @@
...
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
```
wQuelS, line 18: where are your ships supposed to be created and how?

app/main.py
```diff
@@ -1,34 +1,59 @@
...
-        # Find the corresponding deck in the list
-        pass
+        if start[0] != end[0]:
+            for x_coord in range(start[0], end[0] + 1):
+                self.decks.append(Deck(x_coord, start[1]))
+        elif start[1] != end[1]:
+            for y_coord in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], y_coord))
+        else:
+            self.decks.append(Deck(start[0], end[1]))
```
wQuelS, line 27: such logic is better to move to a separate method

app/main.py
```diff
@@ -1,34 +1,59 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
```
wQuelS, line 32: Is it possible that not a valid location is given? If yes - your return type annotation is wrong

app/main.py
```diff
@@ -1,34 +1,59 @@
...
+
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS, line 35: If it is possible for get_deck to return None than it is very unsafe to use is_alive attr on the object that can potentially be None.

app/main.py
```diff
@@ -1,34 +1,59 @@
...
+        drowned = []
+        for deck in self.decks:
+            drowned.append(deck.is_alive)
+        if not any(drowned):
+            self.is_drowned = True
```
wQuelS, line 40: this can be simplified to 1-2 lines by using generator expression and `any`

app/main.py
```diff
@@ -1,34 +1,59 @@
...
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
+        self.field = {}
+        for start, end in ships:
+            ship = Ship(start, end)
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
```
wQuelS, line 50: again, it is better to separate initialization and creation at least in separate methods, that can be called from your `init`

wQuelS (changes requested):

PR #333, opened 2023-10-29
Solution

Daniil-Pankieiev:


OlehShumov (approved):

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.is_drowned = is_drowned
+        self.decks = []
+        for row in range(start[0], end[0] + 1):
+            for column in range(start[1], end[1] + 1):
+                deck = Deck(row, column)
+                self.decks.append(deck)
```
wQuelS, line 25: I think it is better to move such logic to a separate method, that can be called from this `init`

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        current_deck = self.get_deck(row, column)
+        current_deck.is_alive = False
```
wQuelS, line 34: can your `get_deck` potentially return None if there is no ship at that location? If yes, do you think it is safe to use `is_alive` attr from potentially `None` object in such way ?

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        for ship in ships:
+            new_ship = Ship(ship[0], ship[1])
+            for deck in new_ship.decks:
+                self.field[(deck.row, deck.column)] = new_ship
+
+        if not self._validate_field():
+            raise ValueError("Invalid fleet configuration.")
```
wQuelS, line 58: it would be better to create some separate custom errors, to match your specific needs during this task. But overall it is really good that you had an idea to validate data!

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+    def _validate_field(self) -> bool:
+        number_of_ships = {1: 0, 2: 0, 3: 0, 4: 0}
+
+        for ship in set(self.field.values()):
+            number_of_ships[len(ship.decks)] += 1
+        return (sum(number_of_ships.values()) == 10
+                and number_of_ships[1] == 4
+                and number_of_ships[2] == 3
+                and number_of_ships[3] == 2
+                and number_of_ships[4] == 1)
```
wQuelS, line 48: Good, that you have validation of at least a quantity of specific ships. But to be honest, here would also be nice to check if those ships are located at valid location

wQuelS (changes requested):
Nice job, left some comments though

PR #334, opened 2023-10-29
Solution

OlehShumov:


Sel-Fisher (approved):
good code with details!

app/main.py
```diff
@@ -1,34 +1,110 @@
...
+                Deck(start[0], i)
+                for i in range(start[1], end[1] + 1)
+            ]
+        elif end[1] == start[1]:
+            self.decks = [
+                Deck(i, start[1])
+                for i in range(start[0], end[0] + 1)
+            ]
+        else:
+            raise ValueError("Wrong coordinates")
```
wQuelS, line 37: I think such logic should be moved to a separate method. Also, for error handling it would be nice to implement some custom errors that are matching your needs at this task

app/main.py
```diff
@@ -1,34 +1,110 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        self.get_deck(row, column).is_alive = False
```
wQuelS, line 45: can your `get_deck` return None if there is no ship at that location? If yes, do you think it is safe to use `is_alive` attr on an object that potentially can be `None`?

app/main.py
```diff
@@ -1,34 +1,110 @@
...
+        if all([not deck.is_alive for deck in self.decks]):
```
wQuelS, line 46: do you really need that `list comprehension` here? Why not just `generator expression`?

app/main.py
```diff
@@ -1,34 +1,110 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.field = {}
+        for coordinates in ships:
+            ship = Ship(coordinates[0], coordinates[1])
+            for deck in ship.decks:
+                self.field[(deck.row, deck.column)] = ship
```
wQuelS, line 56: again, such logic would better suit a separate method

app/main.py
```diff
@@ -1,34 +1,110 @@
...
+                    if ((i, ship.start[1] - 1) in self.field
+                            or (i, ship.start[1] + 1) in self.field):
+                        return False
+                if ((ship.start[0] - 1, ship.start[1]) in self.field
+                        or (ship.end[0] + 1, ship.end[1]) in self.field):
+                    return False
+
+        if len(ships) == 10 and decks_counting == {1: 4, 2: 3, 3: 2, 4: 1}:
+            return True
+        return False
```
wQuelS, line 110: this can be simplified and decomposed into a few methods that will validate each case separately. But since this is an optional task - fixes are up to you

wQuelS (changes requested):

PR #335, opened 2023-10-29
Battleship

Sel-Fisher:


oleksandr-pupchenko (approved):

app/main.py
```diff
@@ -1,34 +1,60 @@
...
+                 is_drowned: bool = False) -> None:
         # Create decks and save them to a list `self.decks`
-        pass
+        self.is_drowned = is_drowned
+        self.decks = []
+        for i in range((end[0] - start[0]) + (end[1] - start[1]) + 1):
+            if start[0] == end[0]:
+                self.decks.append(Deck(start[0], start[1] + i))
+            elif start[1] == end[1]:
+                self.decks.append(Deck(start[0] + i, start[1]))
```
wQuelS: I think such logic should be moved to a separate method that can be called from this `init`. It is better practice overall

app/main.py
```diff
@@ -1,34 +1,60 @@
...
 
-    def get_deck(self, row, column):
+    def get_deck(self, row: int, column: int) -> Deck | bool:
         # Find the corresponding deck in the list
-        pass
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return False
```
wQuelS: do you really need that `boolean` return here? It can return a `Deck` or `None` and your checks in next methods will still be valid

app/main.py
```diff
@@ -1,34 +1,60 @@
...
 class Battleship:
-    def __init__(self, ships):
+    def __init__(self,
+                 ships: list[tuple[tuple[int, int], tuple[int, int]]]) -> None:
         # Create a dict `self.field`.
         # Its keys are tuples - the coordinates of the non-empty cells,
         # A value for each cell is a reference to the ship
         # which is located in it
-        pass
+        self.field = {ship: Ship(ship[0], ship[1]) for ship in ships}
```
wQuelS: again, I'd suggest to separate initialization and creation of objects

app/main.py
```diff
@@ -1,34 +1,60 @@
...
         # in the ship or not.
-        pass
+        for key in self.field.keys():
+            if (location[0] in range(key[0][0], key[1][0] + 1)
+                    and location[1] in range(key[0][1], key[1][1] + 1)):
+                self.field[key].fire(location[0], location[1])
+                if self.field[key].is_drowned:
+                    return "Sunk!"
+                else:
+                    return "Hit!"
```
wQuelS: do you really need this `else` here?

app/main.py
```diff
@@ -1,34 +1,60 @@
...
-    def fire(self, location: tuple):
+    def fire(self, location: tuple) -> str:
         # This function should check whether the location
         # is a key in the `self.field`
         # If it is, then it should check if this cell is the last alive
         # in the ship or not.
-        pass
+        for key in self.field.keys():
+            if (location[0] in range(key[0][0], key[1][0] + 1)
+                    and location[1] in range(key[0][1], key[1][1] + 1)):
```
wQuelS: this check can be simplified, because you are using dictionaries...

wQuelS (changes requested):

wQuelS (approved):
Looks good to me:)

PR #347, opened 2023-11-13
Solution

loplicat:


app/main.py
```diff
@@ -1,34 +1,80 @@
...
+        ship = self.field[location]
+        ship.fire(*location)
+        if ship.is_drowned:
+            return "Sunk!"
+        return "Hit!"
+
+    def print_field(self) -> None:
+        field = [["~"] * 10] * 10
+        for location, ship in self.field.items():
+            row, column = location
```
Dimosphen1: You could perform unpacking during the cycle declaration as well, there is no need to allocate `location` variable
```suggestion
        for (row, column), ship in self.field.items():
```
Replace your code with this approach.

app/main.py
```diff
@@ -1,34 +1,80 @@
...
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
```
Dimosphen1: For arguments that have default value indicate that they are optional using the type annotation:
```suggestion
            is_drowned: bool | None = False
```

app/main.py
```diff
@@ -1,34 +1,80 @@
...
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
```
Dimosphen1: Indicate tuples of what data types you expect to obtain in type annotation, it is better to be more explicit where possible.

Dimosphen1 (changes requested):

app/main.py
loplicat: Thank you, I've already changed it.

app/main.py
```diff
@@ -1,34 +1,79 @@
...
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(
+            self,
+            row: int,
+            column: int,
+            is_alive: bool = True
```
Dimosphen1, line 9: You could indicate that the `is_alive` argument is optional as well.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #355, opened 2023-12-13
'Solution_battleship'

vvg04021968:


app/main.py
```diff
@@ -1,34 +1,157 @@
...
+                    spase = [i, k_]
+                    ship_deck.append(spase)
+            ship_deck_all += ship_deck
+
+        for ship in self.field.keys():
+            ship = sorted(ship)
+            y0 = ship[0][0] - 1 if ship[0][0] - 1 >= 0 else 0
+            y1 = ship[1][0] + 1 if ship[1][0] + 1 <= 9 else 9
+            x0 = ship[0][1] - 1 if ship[0][1] - 1 >= 0 else 0
+            x1 = ship[1][1] + 1 if ship[1][1] + 1 <= 9 else 9
```
pavlejviki: The 9 is hardcoded based on the assumption that the game field is a 10x10 grid. This works fine for a standard Battleship game, but it reduces flexibility. If you want to change the size of the game field, you would have to find and update all instances of this hardcoded value.

pavlejviki: You can define a constant or a class attribute (like GRID_SIZE) in the Battleship class. This way, you only need to change the grid size in one place, and it will be updated throughout the code.
Example: GRID_SIZE = 10 and then use GRID_SIZE - 1 instead of 9.

app/main.py
```diff
@@ -1,34 +1,157 @@
...
+            for i in range(ship[0][0], ship[1][0] + 1):
+                for k_ in range(ship[0][1], ship[1][1] + 1):
+                    spase = [i, k_]
+                    ship_deck.append(spase)
+            ship_deck_all += ship_deck
+
+        for ship in self.field.keys():
+            ship = sorted(ship)
+            y0 = ship[0][0] - 1 if ship[0][0] - 1 >= 0 else 0
+            y1 = ship[1][0] + 1 if ship[1][0] + 1 <= 9 else 9
```
pavlejviki:
```suggestion
            y1 = min(ship[1][0] + 1, GRID_SIZE - 1)

```

app/main.py
```diff
@@ -1,34 +1,157 @@
...
+        if ships_count != {1: 4, 2: 3, 3: 2, 4: 1}:
+            raise ValidationInputDataError(
+                """There should be 4 single-deck ships
+                There should be 3 double-deck ships
+                There should be 2 three-deck ships
+                There should be 1 four-deck ship"""
+            )
+
+        ship_deck_all = []
+        for ship in self.field.keys():
```
pavlejviki: when iterating over a dictionary, iterating directly over the dictionary itself (like for ship in self.field) is the same as iterating over the dictionary's keys (like for ship in self.field.keys()). So using keys() is redundant

app/main.py
```diff
@@ -1,34 +1,157 @@
...
+            ship = sorted(ship)
+            ship_deck = []
+            for i in range(ship[0][0], ship[1][0] + 1):
+                for k_ in range(ship[0][1], ship[1][1] + 1):
```
pavlejviki: use more descriptive variable keys. Avoid using i,k etc.

app/main.py
pavlejviki: you can simplify the code using min()

app/main.py
```diff
@@ -1,34 +1,157 @@
...
+                    ship_deck.append(spase)
+            ship_deck_all += ship_deck
+
+        for ship in self.field.keys():
+            ship = sorted(ship)
+            y0 = ship[0][0] - 1 if ship[0][0] - 1 >= 0 else 0
+            y1 = ship[1][0] + 1 if ship[1][0] + 1 <= 9 else 9
+            x0 = ship[0][1] - 1 if ship[0][1] - 1 >= 0 else 0
+            x1 = ship[1][1] + 1 if ship[1][1] + 1 <= 9 else 9
+            ship_spase = []
```
pavlejviki, line 113:
```suggestion
            ship_space = []
```

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,34 +1,160 @@
...
+                    )
+
+    def fire(self, location: tuple) -> str:
+        for coords, ship in self.field.items():
+            ship_decks = ship.get_deck(location[0], location[1])
+            if ship_decks is not None:
+                return ship.fire(location[0], location[1])
+        return "Miss!"
+
+    def print_field(self) -> None:
```
Polyakiv-Andrey, line 142: Triple loops are not allowed, think about how to rewrite the code with less nesting

Polyakiv-Andrey (changes requested):

pavlejviki (approved):
Well done!

PR #359, opened 2023-12-26
'Solution'

struhanets:


app/main.py
```diff
@@ -1,34 +1,122 @@
...
+
+        counter = {1: 0, 2: 0, 3: 0, 4: 0}
+        for ship in self.ships:
+            for deck in ship.decks:
+                if deck.row == ship.decks[0] and \
+                        deck.column == ship.decks[0].column:
+                    raise ValueError("This position is wrong to create a ship")
+
+            ship_length = len(ship.decks)
+            if ship_length == 4:
```
vasylhnatiuk: try to avoid duplicating

app/main.py
```diff
@@ -1,34 +1,122 @@
...
+                counter[2] += 1
+            else:
+                counter[1] += 1
+        print(f"we have {counter[4]} four-decks ships")
+        print(f"we have {counter[3]} three-decks ships")
+        print(f"we have {counter[2]} two-decks ships")
+        print(f"we have {counter[1]} one-decks ships")
+
+    def fire(self, location: tuple) -> str:
+        # decks = self.field.items()
```
vasylhnatiuk: delete useless coment

app/main.py
```diff
@@ -1,34 +1,122 @@
...
-        # which is located in it
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
```
vasylhnatiuk: list of what?

app/main.py
```diff
@@ -1,34 +1,122 @@
...
+        self.column = column
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self,
+            start: tuple,
```
```suggestion
            start: tuple[int],
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,123 @@
...
+        self.is_drowned = is_drowned
+        self.decks = [Deck(row, column) for row in
+                      range(self.start[0], self.end[0] + 1)
+                      for column in range(self.start[1], self.end[1] + 1)]
+
+    def get_deck(self, row: int, column: int) -> object:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
```
ArsenPidhoretskyi: This line is redundant

app/main.py
```diff
@@ -1,34 +1,123 @@
...
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = [Deck(row, column) for row in
+                      range(self.start[0], self.end[0] + 1)
+                      for column in range(self.start[1], self.end[1] + 1)]
+
+    def get_deck(self, row: int, column: int) -> object:
```
ArsenPidhoretskyi: specify the correct type which it returns, it returns Optional[Deck]

app/main.py
```diff
@@ -1,34 +1,123 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def generate_decks(self) -> list:
```
ArsenPidhoretskyi, line 30: This method is redundant

app/main.py
```diff
@@ -1,34 +1,123 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        attacked_ship = self.generate_decks()
+        for first, last in attacked_ship:
+            if row == first[0] and column == last[1]:
+                Deck(row, column, is_alive=False)
```
ArsenPidhoretskyi: Just iterate via self.decks

app/main.py
```diff
@@ -1,34 +1,123 @@
...
+        ships = self.ships
+        for ship in ships:
+            for deck in ship.decks:
+                if deck.row == location[0] and deck.column == location[1]:
+                    deck.is_alive = False
+                    if all(not d.is_alive for d in ship.decks):
+                        ship.is_drowned = True
+                        return "Sunk!"
+                    return "Hit!"
+        return "Miss!"
```
ArsenPidhoretskyi: Use location in self.fields
Also reuse ship methods

ArsenPidhoretskyi (changes requested):

app/main.py
struhanets: I do't want to change this method, because it's most clearly for me.

app/main.py
struhanets: without this method, my code would be crashed, because i use it below
`                if location in self.field:
                    ship = self.field[location]
                    deck = ship.get_deck(location[0], location[1])
                    if not deck.is_alive and ship.is_drowned:
                        field_representation += "X"
                    elif not deck.is_alive and not ship.is_drowned:
                        field_representation += "*"
`

ArsenPidhoretskyi: In this case, change logic. This method is totally useless

app/main.py
```diff
@@ -1,34 +1,121 @@
...
+        return ship
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        for first, last in self.decks:
+            if row == first[0] and column == last[1]:
+                Deck(row, column, is_alive=False)
```
```suggestion
        for deck in self.decks:
            if (deck.row, deck.column) == (row, column):
                deck.is_alive = False
```

app/main.py
ArsenPidhoretskyi: It has bad code practices, you do not use ship.fire method
More simple and clear code:
Also update fire method into ship, it must have logic about is ship is_drowned
```
if location in self.fields:
     ship = self.fields[location]
     ship.fire()
     if ship.is_alive:
           return "Hit!"
     return "Sunk!"
return "Miss!"
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #370, opened 2024-01-16
Solution

Zonda001:


app/main.py
```diff
@@ -1,34 +1,69 @@
...
+                 is_drowned: bool = False) -> None:
+        self.decks = [Deck(row, col) for row, col in
+                      self.get_coordinates(start, end)]
+        self.is_drowned = is_drowned
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    @staticmethod
+    def get_coordinates(start: tuple, end: tuple) -> list[tuple]:
```
ArsenPidhoretskyi: This function is redundant, do this in init

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        return [(start[0] + i, start[1] + j) for i
+                in range(end[0] - start[0] + 1) for j in
+                range(end[1] - start[1] + 1)]
```
```suggestion
        return [(start[0] + i, start[1] + j) 
                for i in range(end[0] - start[0] + 1) 
                for j in range(end[1] - start[1] + 1)]
```

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        return next((deck for deck in self.decks if deck.row == row
+                     and deck.column == column), None)
+
+    def fire(self, row: int, column: int) -> None:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            if all(not deck_.is_alive for deck_ in self.decks):
+                self.is_drowned = True
```
```suggestion
            self.is_drowned = all(not deck_.is_alive for deck_ in self.decks):
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #383, opened 2024-02-04
Implemented py-battleship

Aeros72:


app/main.py
```diff
@@ -1,34 +1,99 @@
...
-        pass
+        for row in rows_range:
+            for column in col_range:
+                self.decks.append(Deck(row, column))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def get_deck(self, row: int, column: int) -> Union[Deck, None]:
```
Dimosphen1: Change your return type annotation, instead of `Union` use `|`:
```suggestion
    def get_deck(self, row: int, column: int) -> Deck | None:
```
In case you face errors, use `from __future__ import annotations` import

app/main.py
```diff
@@ -1,34 +1,99 @@
...
+        if location not in self.field:
+            return "Miss!"
+
+        ship = self.field[location]
+        ship.fire(location[0], location[1])
+
+        if ship.is_drowned:
+            return "Sunk!"
+        else:
+            return "Hit!"
```
Dimosphen1: There is no need for `else`, the code will not execute after `return` anyway, so remove it.

Dimosphen1 (changes requested):
Several minor changes were requested.

app/main.py
```diff
@@ -1,34 +1,99 @@
...
+                self.decks.append(Deck(row, column))
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
```
viktoria-rybenchuk: The return None statement outside the loop is not strictly necessary it serves as a default return value.

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #400, opened 2024-02-22
Solution

mykytaso:


app/main.py
```diff
@@ -1,34 +1,51 @@
...
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], column))
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        elif start[1] == end[1]:
+            for row in range(start[0], end[0] + 1):
+                self.decks.append(Deck(row, start[1]))
```
ArsenPidhoretskyi: Use loop into loop inside conditions

app/main.py
```diff
@@ -1,34 +1,51 @@
...
+
+    def get_deck(self, row: int, column: int) -> Deck:
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #401, opened 2024-02-22
Solution

Vasavik-ua:


app/main.py
```diff
@@ -1,34 +1,92 @@
...
+            start: tuple,
+            end: tuple,
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+        if self.start[0] == self.end[0]:
+            for deck in range(self.start[1], self.end[1] + 1):
```
vasylhnatiuk: try ro avoid 2 loops

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,94 @@
...
+        if self.start[0] == self.end[0]:
+            row = False
+            fix_pos = self.end[0]
+            iter_pos = self.end[1]
+            st = self.start[1]
+        for deck in range(st, iter_pos + 1):
+            if not row:
+                self.decks.append(Deck(fix_pos, deck))
+            else:
+                self.decks.append(Deck(deck, fix_pos))
```
ArsenPidhoretskyi: This code can be simplified to:
```
    for column in range(start[1], end[1] + 1):
         self.decks.append(Deck(row, column))
```

app/main.py
```diff
@@ -1,34 +1,94 @@
...
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+        self,
+        start: tuple,
+        end: tuple,
+        is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
```
ArsenPidhoretskyi: No need to save start and end

app/main.py
```diff
@@ -1,34 +1,94 @@
...
+                self.decks.append(Deck(deck, fix_pos))
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(
+        self,
+        row: int,
+        column: int
+    ) -> Deck:
```
ArsenPidhoretskyi, line 30: It can return None also

app/main.py
```diff
@@ -1,34 +1,94 @@
...
+    def fire(
+        self,
+        row: int,
+        column: int
+    ) -> None:
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
+        ship_status = [cell.is_alive for cell in self.decks]
+        if not any(ship_status):
+            self.is_drowned = True
```
```suggestion
        self.is_drowned = not any([cell.is_alive for cell in self.decks])
```

app/main.py
```diff
@@ -1,34 +1,94 @@
...
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
+        self.field = [[None for _ in range(10)] for _ in range(10)]
+        self.ships = [Ship(value[0], value[1]) for value in self.ships]
```
ArsenPidhoretskyi: Wrong realization due to task requirements:

> Create a dict `self.field`.
> Its keys are tuples - the coordinates of the non-empty cells,
> A value for each cell is a reference to the ship

app/main.py
```diff
@@ -1,34 +1,94 @@
...
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
```
ArsenPidhoretskyi: This line is redundant

app/main.py
ArsenPidhoretskyi: No need to save ships

app/main.py
```diff
@@ -1,34 +1,94 @@
...
+        self.field = [[None for _ in range(10)] for _ in range(10)]
+        self.ships = [Ship(value[0], value[1]) for value in self.ships]
+        for ship in self.ships:
+            for cell in ship.decks:
+                self.field[cell.row][cell.column] = ship
+
+    def fire(self, location: tuple) -> str:
+        field_loc = self.field[location[0]][location[1]]
```
```suggestion
        ship = self.field[location[0]][location[1]]
```
Rename this variable

ArsenPidhoretskyi (changes requested):

app/main.py
Vasavik-ua: instance decks of class Ship cannot have None type inside.

vasylhnatiuk (approved):

PR #411, opened 2024-03-13
'Solution'

KravchenkoSerhiii:


app/main.py
```diff
@@ -1,34 +1,82 @@
...
+    ) -> None:
+        self.decks: List[Deck] = []
+        self.is_drowned = is_drowned
+        horizontal = start[0] == end[0]
+        if horizontal:
+            for column in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], column))
+        else:
+            for row in range(start[0], end[0] + 1):
+                self.decks.append(Deck(row, start[1]))
```
ArsenPidhoretskyi: Just use loop into loop instead of code duplicate

app/main.py
```diff
@@ -1,34 +1,82 @@
...
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+            is_drowned: bool = False
+    ) -> None:
+        self.decks: List[Deck] = []
+        self.is_drowned = is_drowned
+        horizontal = start[0] == end[0]
+        row_range = range(start[0], end[0] + 1) \
+            if not horizontal else [start[0]]
+        col_range = range(start[1], end[1] + 1) if horizontal else [start[1]]
+
+        # Use nested loops to iterate over both rows and columns
```
Dimosphen1: Remove this comment, there is no need for it

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+            self,
+            start: Tuple[int, int],
+            end: Tuple[int, int],
+            is_drowned: bool = False
+    ) -> None:
+        self.decks: List[Deck] = []
+        self.is_drowned = is_drowned
+        horizontal = start[0] == end[0]
+        row_range = range(start[0], end[0] + 1) \
+            if not horizontal else [start[0]]
```
Dimosphen1: Avoid using line breakers in your code, make a multi-line if condition or place it inside `()`

app/main.py
```diff
@@ -1,34 +1,84 @@
...
+
+        # Use nested loops to iterate over both rows and columns
+        for row in row_range:
+            for column in col_range:
+                self.decks.append(Deck(row, column))
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> None:
```
Dimosphen1: Return type annotation is not accurate, `deck` cloud be returned as well

Dimosphen1 (changes requested):
Several changes were requested

vasylhnatiuk (approved):

PR #423, opened 2024-04-09
Solution

halyna-baklanova:


app/main.py
```diff
@@ -1,34 +1,56 @@
...
-        # which is located in it
-        pass
+    def __init__(self, ships: tuple) -> None:
+        self.field = {}
+        self.ships = ships
+        for ship in ships:
+            battle_ship = Ship(start=ship[0], end=ship[1])
+            for deck in battle_ship.decks:
+                key = (deck.row, deck.column)
+                self.field[key] = battle_ship
```
pavlopro: Directly assign `(deck.row, deck.column)` into `self.field` as key

app/main.py
```diff
@@ -1,34 +1,56 @@
...
 class Battleship:
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: tuple) -> None:
+        self.field = {}
+        self.ships = ships
```
pavlopro, line 48: Since you don't use (self.ships) anywhere except for creating Ship objects, there no need to alocate variable for it.
Instead just iterate over ships with unpacking:
`for start, end in ships:`
 ...

app/main.py
```diff
@@ -1,34 +1,56 @@
...
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> None:
+        ruined_deck = self.get_deck(row, column)
+        ruined_deck.is_alive = False
+        deck_status = [deck.is_alive for deck in self.decks]
```
pavlopro: Simplify it into one line with line below. Use generator instead of list comprehension

pavlopro (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

pavlopro (changes requested):
Tests Failed

pavlopro (approved):
Good Job!

PR #429, opened 2024-04-13
'Solution'

o-zinevych:


app/main.py
```diff
@@ -1,34 +1,85 @@
...
+                row_deck = [start[0]] * (column_distance - 1)
+            column_deck = []
+            if column_distance >= 2:
+                for i in range(1, column_distance):
+                    column_deck.append(start[1] + i)
+            elif row_distance >= 2:
+                column_deck = [start[1]] * (row_distance - 1)
+            deck_coords = [(row_deck[i], column_deck[i])
+                           for i in range(len(row_deck)) if row_deck]
+            self.decks.extend([Deck(*deck) for deck in deck_coords])
```
ArsenPidhoretskyi: It is so hard code...
Use loop into loop

app/main.py
```diff
@@ -1,34 +1,85 @@
...
+        self.is_drowned = is_drowned
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck:
```

app/main.py
```diff
@@ -1,34 +1,85 @@
...
+    def fire(self, row: int, column: int) -> None:
+        hit_decks_count = 0
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
+        for deck in self.decks:
+            if deck.is_alive:
+                break
+            hit_decks_count += 1
+        if hit_decks_count == len(self.decks):
+            self.is_drowned = True
```
ArsenPidhoretskyi: Use any() instead of count decks

app/main.py
```diff
@@ -1,34 +1,85 @@
...
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.field = {}
+        for ship in ships:
+            ship = Ship(*ship)
+            ship_dict = {(ship.decks[i].row, ship.decks[i].column): ship
+                         for i in range(len(ship.decks))}
```
ArsenPidhoretskyi: Iterate by value instead of index

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #430, opened 2024-04-15
solution

nrjss67:


app/main.py
```diff
@@ -1,34 +1,120 @@
...
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        if start[0] != end[0]:
+            for row in range(start[0], end[0] + 1):
+                self.decks[(row, start[1])] = Deck(row, start[1])
+        else:
+            for column in range(start[1], end[1] + 1):
+                self.decks[(start[0], column)] = Deck(start[0], column)
```
ArsenPidhoretskyi: You can use loop inside loop
```suggestion
        for row in range(start[0], end[0] + 1):
            for column in range(start[1], end[1] + 1):
                self.decks[(start[0], column)] = Deck(start[0], column)
```

app/main.py
```diff
@@ -1,34 +1,120 @@
...
+
+    def get_deck(self, row: int, column: int) -> Callable:
+        return self.fire(self.decks.get((row, column)))
```
ArsenPidhoretskyi, line 32: Why you make fire in deck?

ArsenPidhoretskyi (changes requested):

app/main.py
nrjss67: fixed

app/main.py
nrjss67: I haven't found a more logical use for this method. 
But since it is in the template, I thought it would be correct to pass the deck in this method to the method that already sets the Hit or Sunk status.

vasylhnatiuk (approved):

PR #432, opened 2024-04-15
solution

alonasorochynska:


app/main.py
```diff
@@ -1,10 +1,19 @@
+from typing import Tuple
+
+
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    def __init__(self, row: int, column: int, is_alive: bool = True) -> None:
```
Oleksl888: bad formatting here

app/main.py
```diff
@@ -1,10 +1,19 @@
...
+        self.row = row
+        self.column = column
+        self.is_alive = is_alive
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
+    def __init__(self, start: Tuple[int, int], end: Tuple[int, int], is_drowned: bool = False) -> None:
```
Oleksl888: here also - use one line for each param

Oleksl888 (changes requested):
Not implemented

app/main.py
```diff
@@ -1,34 +1,70 @@
...
+            start: Tuple[int, int],
+            end: Tuple[int, int],
+            is_drowned: bool = False
+    ) -> None:
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = []
+
+        if start[1] != end[1]:
```
Oleksl888, line 28: you should move this to a separate method e.g. create_ship_decks that takes start and end and call it in __init__

app/main.py
```diff
@@ -1,34 +1,70 @@
...
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+
+        for ship in ships:
```
Oleksl888, line 51: move this to a separate method and then call it on __init__

app/main.py
```diff
@@ -1,34 +1,70 @@
...
+                self.field[ship[0]] = unit
+
+    def fire(self, location: tuple) -> str:
+        if location not in self.field:
+            return "Miss!"
+        if location in self.field:
+            self.field[location].fire(location[0], location[1])
+            if self.field[location].is_drowned:
+                return "Sunk!"
+            else:
```
Oleksl888, line 69: else is unnecessary here

Oleksl888 (changes requested):

PR #436, opened 2024-04-22
Done

andriy-py:


app/main.py
```diff
@@ -1,34 +1,73 @@
...
+            start, end = ship
+            new_ship = Ship(start, end)
+            for deck in new_ship.decks:
+                self.field[(deck.row, deck.col)] = new_ship
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field:
+            ship = self.field[location]
+            return ship.fire(*location)
+        else:
```
vasylhnatiuk: you can avoid else here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            if all(not deck.is_alive for deck in self.decks):
+                self.is_drowned = True
+                return "Sunk!"
+            else:
+                return "Hit!"
+        else:
+            return "Miss!"
```
```suggestion
                return "Sunk!"
            return "Hit!"
        return "Miss!"
```

app/main.py
```diff
@@ -1,34 +1,73 @@
...
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {}
+        self.ships = ships
+        for ship in ships:
+            start, end = ship
+            new_ship = Ship(start, end)
+            for deck in new_ship.decks:
+                self.field[(deck.row, deck.col)] = new_ship
```
vsmutok: try to use dict comprehensions here

vsmutok (changes requested):

app/main.py
```diff
@@ -1,34 +1,68 @@
...
-        # Find the corresponding deck in the list
-        pass
+        self.is_drowned = is_drowned
+        self.decks = []
+        if start[0] == end[0]:
+            for col in range(start[1], end[1] + 1):
+                self.decks.append(Deck(start[0], col))
+        elif start[1] == end[1]:
+            for row in range(start[0], end[0] + 1):
+                self.decks.append(Deck(row, start[1]))
```
ArsenPidhoretskyi, line 19: You can use loop into loop instead of if statements
```suggestion
        for col in range(start[1], end[1] + 1):
            for row in range(start[0], end[0] + 1):
                self.decks.append(Deck(row, col))
```

ArsenPidhoretskyi (changes requested):

PR #448, opened 2024-06-05
solution

AndrewwSSS:


app/main.py
```diff
@@ -1,34 +1,96 @@
...
+
+        attacked_deck = filtered[0]
+        if not attacked_deck.is_alive:
+            return "Miss!"
+        else:
+            self.field[attacked_deck].fire(location[0], location[1])
+            if self.field[attacked_deck].is_drowned:
+                return "Sunk!"
+            else:
+                return "Hit!"
```
DevSpaciX: `Else` statement is redundat here , cuz you call `return`

app/main.py
```diff
@@ -1,34 +1,96 @@
...
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        decks = list(
+            filter(lambda deck: (deck.row == row
+                                 and deck.column == column),
+                   self.decks)
+        )
+        if decks:
+            return decks[0]
```
DevSpaciX: Try to use this` :=` operator , in special cases like this it can be really usefull
```
if decks := `your lamda here` :
    return decs[0]
```

look where it can be used too

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #451, opened 2024-06-07
py-battleship v1

BornToLivee:


app/main.py
```diff
@@ -1,34 +1,65 @@
...
+                    if not deck.is_alive and ship.is_drowned:
+                        print("x   ", end=" ")
+                    elif not deck.is_alive:
+                        print("*   ", end=" ")
+                    else:
+                        print(u"\u25A1   ", end=" ")
+                else:
+                    print("~   ", end=" ")
+            print("")
+            print()
```
DevSpaciX: Is this print is necessery ?) You just printing .. Nothing ?)

DevSpaciX (changes requested):

app/main.py
BornToLivee: i dont remember my logic, maybe i just wanted to print extra space or smth :)

DevSpaciX (approved):

PR #464, opened 2024-07-27
Implemented solution

dimak20:


app/main.py
```diff
@@ -1,34 +1,65 @@
...
 class Battleship:
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list) -> None:
+        self.field = {
+            ship: Ship(ship[0], ship[1]) for ship in ships
```
MNDonut: you can use unpacking here

app/main.py
```diff
@@ -1,34 +1,65 @@
...
 
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def fire(self, location: tuple) -> str:
+        for ship in self.field:
+            attempt = self.field[ship].fire(location[0], location[1])
```
MNDonut: you can use unpacking here

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+            ship: Ship(ship[0], ship[1]) for ship in ships
+        }
 
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def fire(self, location: tuple) -> str:
```
MNDonut: specify what type of data this tuple contains

app/main.py
```diff
@@ -1,34 +1,65 @@
...
 
 
 class Battleship:
-    def __init__(self, ships):
-        # Create a dict `self.field`.
-        # Its keys are tuples - the coordinates of the non-empty cells,
-        # A value for each cell is a reference to the ship
-        # which is located in it
-        pass
+    def __init__(self, ships: list) -> None:
```
MNDonut: what type of data this list contains?

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(
+            self,
+            row: int,
+            column: int
+    ) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row:
+                if deck.column == column:
```
MNDonut: you can use "and" operator here

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(
+            self,
+            row: int,
+            column: int
+    ) -> int:
+        got_deck = self.get_deck(row, column)
+        if got_deck is None or not got_deck.is_alive:
```
MNDonut: just "if got_deck" is enough

app/main.py
```diff
@@ -1,34 +1,65 @@
...
+            return 1
+        got_deck.is_alive = False
+        if all(not deck.is_alive for deck in self.decks):
+            self.is_drowned = True
+            return 3
+        return 2
```
MNDonut: isn't it better to return string Miss, etc instead of unknown numbers?

app/main.py
```diff
@@ -1,34 +1,65 @@
...
-        # in the ship or not.
-        pass
+    def fire(self, location: tuple) -> str:
+        for ship in self.field:
+            attempt = self.field[ship].fire(location[0], location[1])
+            if attempt >= 2:
+                if attempt == 3:
+                    return "Sunk!"
+                return "Hit!"
+        return "Miss!"
```
MNDonut: A bit strange logic as for a dev who sees your code for the first time. You can return this string in the fire method of Ship class

MNDonut (changes requested):

app/main.py
dimak20: image

dimak20: I've just reread comments in main branch
Ship:
 def fire(self, row, column):
        pass
Battleship:
def fire(self, location: tuple):
        # in the ship or not.
        pass
 I think I should return string in Battleship fire method and just override attributes in Ship fire method

MNDonut (approved):

PR #468, opened 2024-08-01
The solution of the task

Ja0ni:


app/main.py
```diff
@@ -1,34 +1,129 @@
...
+        self.decks = []
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return
```
viktoria-rybenchuk: why do you need this return?

viktoria-rybenchuk (changes requested):

pavlopro (approved):
Good Job!

PR #470, opened 2024-08-07
Solution

OleksiukStepan:
Added and implemented:

1. Added the **Cell** class for filling empty cells and as a parent for Deck.
2. Added a special attribute to the **Deck** class to customize its symbol on the field ("□")
3. class **Ship** methods:

- get_deck  _(Return the deck located at the coordinates)_
- fire  _(Match Deck as damaged ("*") or drowned ("x") if all ship Decks are damaged)_
- if_all_deck_is_damaged  _(Verifies if all decks are damaged)_
- create_decs  _(Create Decks based on ship length)_
- create_ship  _(Instantiate Ship)_

4.  class **Battleship** methods:

- getter, setter ship  _(Also for deploying ships to the field)_
- fire  _(Return information about the shot. Also can display field changes)_
- add_ships_to_the_field  _(Placing ships on the field based on their coordinates)_
- print_field  _(Printing field to the console)_
- get_symbol_from_cell  _(Retrieving symbol ("□", "~", "*", "x") from the cell)_
- create_empty_field  _(Create a field with the required coordinates and fill it with Cells ("~"))_


app/main.py
```diff
@@ -1,34 +1,127 @@
...
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+        return list_of_decks
+
+    @classmethod
+    def create_ship(cls, ship_cords: tuple) -> Ship:
+        return Ship(*ship_cords)
```
MNDonut: use cls instead of Ship, it's the purpose of classmethod

app/main.py
```diff
@@ -1,34 +1,127 @@
...
+                print()
+            print(self.get_symbol_from_cell(cords, cell), end="  ")
+        print("")
+
+    @staticmethod
+    def get_symbol_from_cell(cords: tuple[int, int], cell: Cell | Ship) -> str:
+        if isinstance(cell, Ship):
+            return cell.get_deck(cords[0], cords[1]).symbol
+        else:
+            return cell.symbol
```
MNDonut: no need for else statement because if statement has return

app/main.py
```diff
@@ -1,34 +1,127 @@
...
+        return deck[0] if deck else None
+
+    def fire(self, row: int, column: int) -> str:
+        if self.is_drowned:
+            return "The ship in this location already drowned"
+
+        deck = self.get_deck(row, column)
+        deck.is_alive = False
+        deck.symbol = "*"
+        return "Sunk!" if self.if_all_deck_is_damaged() else "Hit!"
```
MNDonut: a bit don't get it. In case deck is not found, get_deck returns None, but you set deck.is_alive = False anyway. Shouldn't it be missed?

MNDonut (changes requested):

app/main.py
OleksiukStepan: You're correct. The method get_deck cannot return None, as each ship consists solely of decks; thus, it will always return a deck. I've made some adjustments to ensure it functions more accurately.

OleksiukStepan:
Small changes:

-  Ship -> cls
- removed else

Updated get_deck logic:

- Now get_deck returns only the deck. 
- Added a check for duplicate coordinates: if the deck is already hit, print a message.

MNDonut (approved):

PR #471, opened 2024-08-12
Solution

DSahalatyi:


app/main.py
```diff
@@ -1,34 +1,91 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+        return None
+
+    def get_decks(self) -> list:
+        return self.decks
+
+    def fire(self, row: int, column: int) -> bool:
+        self.get_deck(row, column).is_alive = False
```
MNDonut, line 26: get_deck may return None and error will be raised here

MNDonut (approved):

PR #473, opened 2024-08-13
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,34 +1,109 @@
...
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple[tuple, tuple]]) -> None:
+        self.field = {}
+        for ship in ships:
+            current_ship = Ship(ship[0], ship[1])
```
MNDonut: you can use unpacking here

app/main.py
```diff
@@ -1,34 +1,109 @@
...
+            if deck.row == row and deck.column == column:
+                return deck
 
-    def fire(self, row, column):
-        # Change the `is_alive` status of the deck
-        # And update the `is_drowned` value if it's needed
-        pass
+    def fire(self, row: int, column: int) -> bool:
+        deck = self.get_deck(row, column)
+        if not self.is_drowned and deck.is_alive:
```
MNDonut: let's swap conditions because it's easier to read the condition without NOT first

app/main.py
```diff
@@ -1,34 +1,109 @@
...
+
+        for row in empty:
+            filled_field += "\t".join(row) + "\n"
+
+        print(filled_field)
+
+    def _validate_ships_count(self) -> bool:
+        ships_count = [0, 0, 0, 0]
+        for ship in set(self.field.values()):
+            ships_count[len(ship.decks) - 1] += 1
```
MNDonut: Do you call these methods somewhere? You had to implement and use them

app/main.py
```diff
@@ -1,34 +1,109 @@
...
+            (-1, 1), (1, -1), (1, 0), (0, 1),
+        ]
+
+        for (row, column), ship in self.field.items():
+            for dr, dc in directions:
+                adjacent = (row + dr, column + dc)
+                if adjacent in self.field and self.field[adjacent] != ship:
+                    return False
+
+        return False
```
MNDonut: This method never returns True for 63 line. It means it always returns False and "if not self._validate..." always is True and raises error in line 60-65

app/main.py
```diff
@@ -1,34 +1,109 @@
...
+
+        return "Miss!"
+
+    def _validate_field(self) -> None:
+        ships = set(self.field.values())
+        if not (
+            len(ships) == 10
+            or self._validate_ships_count
+            or self._validate_ship_location
+        ):
```
MNDonut: shouldn't it in vice-versa way? LIke if len != 10 or self._validate returns False(meaning not correct), then error should be raised?

MNDonut (changes requested):

MNDonut (approved):

PR #476, opened 2024-08-19
Resolve

ihorutkin:


app/main.py
```diff
@@ -1,34 +1,69 @@
+from __future__ import annotations
```
MNDonut: why do you need this? Seems like you use ordinary types in annotation

app/main.py
```diff
@@ -1,34 +1,69 @@
...
-        # in the ship or not.
-        pass
+
+    def __init__(self, ships: list) -> None:
+        self.ships = ships
+        self.field = {}
+
+        for ship in ships:
+            ship_instance = Ship(ship[0], ship[1])
+            for deck in Ship(ship[0], ship[1]).decks:
```
MNDonut: 1. You can use unpacking here
2. In the loop use ship_instance variable

app/main.py
```diff
@@ -1,34 +1,69 @@
...
-        pass
-
-    def fire(self, location: tuple):
-        # This function should check whether the location
-        # is a key in the `self.field`
-        # If it is, then it should check if this cell is the last alive
-        # in the ship or not.
-        pass
+
+    def __init__(self, ships: list) -> None:
```
MNDonut: specify what data this list contains

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        self.ships = ships
+        self.field = {}
+
+        for ship in ships:
+            ship_instance = Ship(ship[0], ship[1])
+            for deck in Ship(ship[0], ship[1]).decks:
+                self.field[(deck.row, deck.column)] = ship_instance
+
+    def fire(self, location: tuple) -> str:
```
MNDonut: specify what data this tuple contains

app/main.py
```diff
@@ -1,34 +1,69 @@
...
+        if location in self.field:
+            ship = self.field[location]
+            ship.fire(*location)
+
+            if ship.is_drowned:
+                return "Sunk!"
+            else:
+                return "Hit!"
+        else:
```
MNDonut: No need for else because if statement exactly returns something as there is another else statement

MNDonut (changes requested):

app/main.py
ihorutkin: Not, get_deck function  returns Deck object, but it only one place where I using it, so would it be right if I deleted this and just wrote “Deck”, it also works

LiudmylaKulinchenko (approved):
Good job 👏

PR #491, opened 2024-10-14
Develop

dmytrik:


app/main.py
```diff
@@ -1,34 +1,214 @@
...
+
+    @staticmethod
+    def create_decks(start: tuple, end: tuple) -> list[Deck]:
+        result = []
+        if start == end:
+            result.append(Deck(start[0], start[1]))
+            return result
+        if start[0] == end[0]:
+            for i in range(start[1], end[1] + 1):
+                result.append(Deck(start[0], i))
```
AnyoneClown: Use more descriptive variable. `column`, for example or `row` in other case

app/main.py
```diff
@@ -1,34 +1,214 @@
...
 
 class Battleship:
-    def __init__(self, ships):
+    def __init__(self, ships: list[tuple]) -> None:
         # Create a dict `self.field`.
         # Its keys are tuples - the coordinates of the non-empty cells,
         # A value for each cell is a reference to the ship
         # which is located in it
-        pass
+        self.ships = {coords: Ship(coords[0], coords[1]) for coords in ships}
```
AnyoneClown, line 111: Why are you using `dict` here? Almost in all other methods you are using `.values()` to get data. So why can't we create simple list?

P.S. I'm not saying that you're wrong, you can prove your point and argue for using `dict`

app/main.py
```diff
@@ -1,34 +1,214 @@
...
+            if not is_found:
+                symbols_list.append("~")
+
+        symbols = ""
+        for index, symbol in enumerate(symbols_list):
+            if index % 10 == 0 and index != 0:
+                symbols += "\n"
+            symbols += f"{symbol}   "
+
+        print(symbols)
```
AnyoneClown: It's hard to read this method. We should simplify it.

1. Let's first create a 10x10 grid initialized with water symbols(`~`)
2. Then fill the grid with ship symbols(+ check their status)
3. Create single string of symbols

app/main.py
```diff
@@ -1,34 +1,214 @@
...
+    def _validate_field(self) -> None:
+        coords = self.ships.keys()
+
+        for x_, y_ in coords:
+            condition = (0 <= x_[0] < 10
+                         and 0 <= x_[1] < 10
+                         and 0 <= y_[0] < 10
+                         and 0 <= y_[1] < 10)
+            if not condition:
+                raise VisibilityError
```
AnyoneClown: Refactor using `all`

app/main.py
```diff
@@ -1,34 +1,214 @@
...
+
+        for ship in ships:
+            if len(ship) == 1:
+                one_decks += 1
+            if len(ship) == 2:
+                two_decks += 1
+            if len(ship) == 3:
+                three_decks += 1
+            if len(ship) == 4:
+                four_decks += 1
```
AnyoneClown: We can create it as a dictionary `{1(first deck): 0, 2(second deck): 0}`

app/main.py
```diff
@@ -1,34 +1,214 @@
...
+        assert one_decks == 4, "There should be 4 single-deck ships"
+        assert two_decks == 3, "There should be 3 double-deck ships"
+        assert three_decks == 2, "There should be 2 three-deck ships"
+        assert four_decks == 1, "There should be 1 four-deck ship"
+
+        invalid_cells = []
+        for coord in coords:
+            cells = self._get_invalid_cells_for_one_ship(coord)
+            invalid_cells += cells
```
AnyoneClown: We can use `set comprehension` here

app/main.py
```diff
@@ -1,34 +1,214 @@
...
+            for deck in decks:
+                if cell[0] == deck.row and cell[1] == deck.column:
+                    raise Exception("ships shouldn't be located in"
+                                    " the neighboring cells"
+                                    " (even if cells are neighbors"
+                                    " by diagonal)")
+
+    def _get_invalid_cells_for_one_ship(
+            self,
+            coord: tuple
```
AnyoneClown, line 206: Add more descriptive annotation, check all `funcs` and `methods`

AnyoneClown (changes requested):

app/main.py
dmytrik: task's condition says to use dictionary and keys help me in method where i find all cells around the ship in which other ships can't be located

app/main.py
```diff
@@ -1,34 +1,230 @@
...
+            if ship.is_drowned:
+                continue
+            deck = ship.get_deck(location[0], location[1])
+            if deck:
+                current_ship = ship
+
+        if current_ship and not current_ship.is_drowned:
+            return "Hit!"
+        if current_ship and current_ship.is_drowned:
+            return "Sunk!"
```
Arsen-hrynevych, line 144: After hitting a ship's deck, you should check whether all decks of that ship are "sunk" (i.e., is_alive = False) to return "Sunk!".

app/main.py
```diff
@@ -1,34 +1,230 @@
...
+
+        for ship in self.ships.values():
+            if ship.is_drowned:
+                continue
+            deck = ship.get_deck(location[0], location[1])
+            if deck:
+                current_ship = ship
+
+        if current_ship and not current_ship.is_drowned:
+            return "Hit!"
```
Arsen-hrynevych, line 142: If other decks of the ship are still alive, return "Hit!".

app/main.py
```diff
@@ -1,34 +1,230 @@
...
+        if current_ship and current_ship.is_drowned:
+            return "Sunk!"
+        return "Miss!"
```
Arsen-hrynevych: If no ship exists at the fired location, return "Miss!"

app/main.py
```diff
@@ -1,34 +1,230 @@
...
+        decks = sum([ship.decks for ship in self.ships.values()], [])
+
+        for deck in decks:
+            row, column = deck.row, deck.column
+            sea[row][column] = deck.symbol
+
+        for row in sea:
+            print("    ".join(row))
+
+    def _validate_field(self) -> None:
```
Arsen-hrynevych, line 161: Ensure that exactly 10 ships exist, with the correct number of ships of different sizes (1, 2, 3, and 4 decks).
No two ships should be adjacent, even diagonally.

app/main.py
```diff
@@ -1,34 +1,230 @@
...
+            self._get_invalid_cells_for_one_ship(coord)
+            for coord in coords
+        ], [])
+
+        invalid_cells = set(invalid_cells)
+        decks = sum([ship.decks for ship in self.ships.values()], [])
+        for cell in invalid_cells:
+            for deck in decks:
+                if cell[0] == deck.row and cell[1] == deck.column:
+                    raise Exception("ships shouldn't be located in"
```
Arsen-hrynevych: You could create your own Exception

Arsen-hrynevych (changes requested):
Mostly correct, just needs a few adjustments and it'll be ok.

app/main.py
dmytrik: I do this check in the __init__ method after creating the self.ships field as specified in the task condition. Logically, if any of the validation conditions is not met, an error will occur and we will not be able to access this method as well as others.

Arsen-hrynevych (approved):
ok

PR #494, opened 2024-10-16
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1,34 +1,112 @@
...
+        for deck in self.decks:
+            if deck.row == row and deck.column == column:
+                return deck
+
+        return None
+
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+
+        if deck:
```
Arsen-hrynevych, line 37: You can simplify the logic by removing the call to get_deck twice.

Arsen-hrynevych (approved):
ok

PR #523, opened 2024-12-30
Solution

ruslanyemets:


app/main.py
```diff
@@ -1,34 +1,172 @@
...
+        self.start = start
+        self.end = end
+        self.is_drowned = is_drowned
+        self.decks = self.create_ship(start, end)
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    @staticmethod
+    def create_ship(start: tuple[int], end: tuple[int]) -> list:
```
L1nk3rrr: return list of Decks
`list[Deck]`

app/main.py
```diff
@@ -1,34 +1,172 @@
...
-        # in the ship or not.
-        pass
+    def __init__(self, ships: list[tuple]) -> None:
+        self.field = self.create_field(ships)
+
+        try:
+            self._validate_field()
+        except ValueError as ve:
+            print(ve)
+            exit()
```
L1nk3rrr: What is the method exit?

app/main.py
```diff
@@ -1,34 +1,172 @@
...
+
+    @staticmethod
+    def create_field(ships: list[tuple]) -> dict:
+        ships_list = []
+        field_dict = {}
+
+        for ship in ships:
```
L1nk3rrr: is it really `ship` or it's a coordinates of ship?

app/main.py
```diff
@@ -1,34 +1,172 @@
...
+            ships_list.append(Ship(ship[0], ship[1]))
+
+        for ship in ships_list:
+            for deck in ship.decks:
+                field_dict[(deck.row, deck.column)] = ship
+
+        return field_dict
+
+    def fire(self, location: tuple) -> str:
+        if location in self.field.keys():
```
L1nk3rrr: you dont need to call `keys()`, __contains__ method in dict already checking for keys

so simply use `if location in self.field`

app/main.py
```diff
@@ -1,34 +1,172 @@
...
+                return "Sunk!"
+            return "Hit!"
+        return "Miss!"
+
+    def print_field(self) -> None:
+        field_string = ""
+
+        for row in range(10):
+            for column in range(10):
+                if (row, column) in self.field.keys():
```
L1nk3rrr: you dont need to call keys(), contains method in dict already checking for keys

so simply use `if (row, column) in self.field`

app/main.py
```diff
@@ -1,34 +1,172 @@
...
+                    ship = self.field[(row, column)]
```
L1nk3rrr: also I see that `(row, column)` used twice, so maybe better to make a variable for it

app/main.py
```diff
@@ -1,34 +1,172 @@
...
+                    field_string += "~" + "\t"
+            field_string += "\n"
+
+        print(field_string)
+
+    def _validate_field(self) -> None:
+        ships_list = []
+
+        for (row, column) in self.field.keys():
+            ship = self.field[(row, column)]
```
L1nk3rrr: Also here you can use `for ship in self.field.values()` to simplify your code

L1nk3rrr: And if you used it, your code can be rewrite to list comprehension

L1nk3rrr (changes requested):
Nice! You did main logic, and extended version! I left some comments for improvement of your code

L1nk3rrr (approved):
Well done!

PR #538, opened 2025-02-16
Solution

Baranotik15:


app/main.py
```diff
@@ -1,34 +1,133 @@
...
+    def fire(self, row: int, column: int) -> str:
+        deck = self.get_deck(row, column)
+        if deck:
+            deck.is_alive = False
+            deck.symbol = "x"
+            if self.is_ship_sunk():
+                self.is_drowned = True
+                return "Sunk!"
+            else:
+                return "Hit!"
```
oliasymak24: You can call `return` just outside the `if` without `else`

app/main.py
```diff
@@ -1,34 +1,133 @@
...
+            (-1, -1),
+            (-1, 1),
+            (1, -1),
+            (1, 1)
+        ]
+        for dr, dc in directions:
+            new_row, new_col = row + dr, column + dc
+            if (0 <= new_row < 10 and 0 <= new_col < 10
+                    and self.field[new_row][new_col] == "□"):
+                return True
```
oliasymak24: Try to use here any()

app/main.py
```diff
@@ -1,34 +1,133 @@
...
+    def fire(self, location: Tuple[int, int]) -> str:
+        row, column = location
+        if 0 <= row < 10 and 0 <= column < 10:
+            hit_result = "Miss!"
+            for ship in self.ships:
+                hit_result = ship.fire(row, column)
+                if hit_result != "Miss!":
+                    break
+            return hit_result
+        return "Invalid coordinates!"
```
oliasymak24: I think in this method  after each hit, self.field needs to be updated to reflect the new game state.

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #541, opened 2025-02-20
Solution

kram3ko:


app/main.py
```diff
@@ -1,34 +1,106 @@
...
+
+
+@dataclass
 class Deck:
-    def __init__(self, row, column, is_alive=True):
-        pass
+    row: int
+    column: int
+    is_alive: bool = True
+    marker = "□"
```
Boyok1337: Why you dont use `print(u"\u25A1")` this element from Readme instruction as provided?

Boyok1337 (approved):
Good one! But do not forget change `marker` and explain it to me

app/main.py
kram3ko: fixed to unicode symbol

kram3ko:
> Good one! But do not forget change `marker` and explain it to me

done. Thank you

PR #549, opened 2025-03-10
Completed task py-battleship

Yurkiv313:


PrimeGlorious (approved):

app/main.py
```diff
@@ -1,34 +1,107 @@
...
+    def __repr__(self) -> str:
+        return f"({self.row}, {self.column})"
 
 
 class Ship:
-    def __init__(self, start, end, is_drowned=False):
-        # Create decks and save them to a list `self.decks`
-        pass
+    def __init__(
+            self, start: tuple, end: tuple, is_drowned: bool = False
```
Boyok1337: Add annotation for element in tuple if possible and use `from typing import List, Tuple, Optional` for it

app/main.py
```diff
@@ -1,34 +1,107 @@
...
+                for row in range(end[0], start[0] + 1):
+                    self.decks.append(Deck(row, col))
+                return
+
+            col = start[1]
+            for row in range(start[0], end[0] + 1):
+                self.decks.append(Deck(row, col))
+            return
+
+    def get_deck(self, row: int, column: int) -> Any:
```
Boyok1337: Here return can be Optional[Deck] or None, but not Any

Boyok1337 (changes requested):
Also, listen to the recommendations of AI

Y-Havryliv (approved):

PR #558, opened 2025-04-06
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
_create_deck_list calculate ship length and create new deck base on it. It works correctly.
add validation of type ship in Battleship __init__
collision validation detect all neighbour cell of deck (all 8 directions included diagonal neighbour)

olek007:
@mateacademy-ai-mentor 
The current implementation do not creates a rectangle of decks instead of a line. No need for separate logic to detect if horizontal or vertical.
The current collision validation logic checks all neighboring cells (including diagonals). Your detects only up, down, left, right collision.

PR #601, opened 2025-09-02
Solution

Konstantin-Svt:


Konstantin-Svt:
@mateacademy-ai-mentor can't find redundant else after a return in app/main.py: print_field.

PR #609, opened 2025-10-18
Solution

Tersting:


Tersting:
Your solution is approved. You implemented a robust Battleship with proper ship construction from endpoints, strict bounds checking, rejection of diagonal ships, overlap prevention, adjacency rules, and persistent hit/sunk state with fire returning only "Miss!", "Hit!", or "Sunk!" as required.

vsmutok (approved):
Great!

PR #680, opened 2026-04-10
create a battleship game

pawunder1992:


app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        if current_deck.is_alive is True:
+            current_deck.is_alive = False
+        result = 0
+        for deck in self.decks:
+            if deck.is_alive:
+                result += 1
+        if result == 0:
+            self.is_drowned = False
+            return "Sunk!"
+        return "Hit!"
```
AnyoneClown, line 47: Why are you saying, that `is_drowned = False`, if it's not alive?

app/main.py
```diff
@@ -1,34 +1,73 @@
...
+        self.is_drowned = is_drowned
+        self.create_deckds()
 
-    def get_deck(self, row, column):
-        # Find the corresponding deck in the list
-        pass
+    def get_deck(self, row: int, column: int) -> Deck | None:
+        for deck in self.decks:
+            if row == deck.row and column == deck.column:
+                return deck
```
AnyoneClown, line 23: Store decks in the dict, in that case u will have O(1) speed for lookup, and then search for a deck with `get()` method.

In you dictionary - you will have this type of key -> `(row,col)` and value is gonna be the instance of your deck

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

